PR #3, opened 2020-05-13
Update yugaware to allow migration to helm3

Arnav15:
Signed-off-by: Arnav Agarwal <14933889+Arnav15@users.noreply.github.com>
Makes the necessary changes to migrate YW from helm2 to helm3. Also adds compatibility in Yugabyte for being able to use a customizable clusterIP since helm3 doesn't work well with charts deployed with an empty clusterIP [https://github.com/helm/helm/issues/6378].
Migration instructions: https://helm.sh/blog/migrate-from-helm-v2-to-helm-v3/
For YW upgrades from a migrated chart:
Just set the clusterIP value in `yugaware.services.clusterIP` and set `helm2Legacy` to true.
For YB upgrades from a migrated chart:
Set the clusterIP value for each `serviceEndpoints` as `clusterIP` and set `helm2Legacy` to true.
Running an upgrade from a helm3 deployment from an older chart will also need the `clusterIP` field to be set


iSignal (approved):

iSignal:
Looks good! Might be a good idea to verify that a helm2 deployment created before this chart is compatible with one created after this chart.

PR #9, opened 2020-06-17
[stable/yugabyte] set workingDir to /mnt/disk0

### LINKED ISSUE #11: Saving core dumps to PersistentVolumes
Usually `kernel.core_pattern` is set to a relative path which is something like `core.%e.%p.%t`. In such scenarios the core dumps are saved in same directory as of process' current working directory (i.e. `/proc/$pid/cwd`).

We can change the working directory of yb-master as well as yb-tserver process by changing container's `workingDir` for a pod. Setting this to a path from PV will make it possible to persist the core dumps across container restarts.

https://v1-16.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#container-v1-core
> `workingDir`: Container's working directory. If not specified, the container runtime's default will be used, which might be configured in the container image. Cannot be updated.

bhavin192:
- If host has kernel.core_pattern set to some relative path like `core.%e.%p.%t`, then the core dumps will get saved in the volume mounted at `/mnt/disk0`

**Note**: after this change, `kubectl exec …` commands will have to use absolute paths to binaries from bin directory, i.e. `/home/yugabyte/bin/ysqlsh` instead of `bin/ysqlsh`.

### Scenarios tested
- Installation with `storage.ephemeral=true` and `storage.ephemeral=false`
- Installation with `tls.enabled=true` and `tls.enabled=false`

### How to check if workingDir is set correctly
- Output of given commands should match
  ```console
  $ kubectl exec -it yb-master-1 -nyb-chart -- bash
  [root@yb-master-1 disk0]# pwd
  /mnt/disk0
  ```

Fixes #11

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -246,6 +246,11 @@ spec:
         {{ else }}
 {{ toYaml $root.Values.resource.tserver | indent 10 }}
         {{ end }}
+        # core dumps are collected to workingDir if
+        # kernel.core_pattern is set to a relative path like
+        # core.%e.%p.%t ref:
+        # https://github.com/yugabyte/charts/issues/11
+        workingDir: "/mnt/disk0"
         command:
         {{ if eq .name "yb-masters" }}
           - "/home/yugabyte/bin/yb-master"


stable/yugabyte/templates/service.yaml
```diff
@@ -246,6 +246,7 @@ spec:
         {{ else }}
 {{ toYaml $root.Values.resource.tserver | indent 10 }}
         {{ end }}
+        workingDir: "/mnt/disk0"
```
iSignal, line 253: Add a comment linking to the issue number and explaining why we did this change.

iSignal (approved):

PR #13, opened 2020-07-09
Add flags to enable TLS

### LINKED ISSUE #4726: Document k8s client usage of yugabyte helm chart with TLS on
The yugabyte helm chart already supports deploying with TLS on using the override tls.enabled as shown here - https://github.com/yugabyte/charts/blob/master/stable/yugabyte/templates/service.yaml#L33

A client should be able to use the yugabyte client cert at https://github.com/yugabyte/charts/blob/master/stable/yugabyte/templates/service.yaml#L8 within their pod using ysqlsh as described at https://docs.yugabyte.com/latest/secure/tls-encryption/connect-to-cluster/#connect-to-a-yugabytedb-cluster (in the non-k8s case)

We need to 
(a) verify that this flow still works and 
(b) document the steps in (a) and add them to the docs at https://docs.yugabyte.com/latest/deploy/kubernetes/clients/

@Arnav15 @rkarthik007

baba230896:
- Add flags to enable TLS

**Edited:**

> ~~**Tested Scenarios:**~~
> 
> ~~1. insecure: false, insecureClient: false - Client with Certificates are only allowed~~
> ~~2. insecure: false, insecureClient: true - All clients are allowed (With or Without certificates).~~
> ~~3. insecure: true, insecureClient: false - Client with Certificates are only allowed~~
> ~~4. insecure: true, insecureClient: true - All clients are allowed.~~
> 
> The `insecureClient` has been removed. Use `insecure` to toggle the insecure communication.

- Modified the bind-address of pgsql_proxy_bind_address from `{{ printf "$(POD_IP)" }}:5433` to `0.0.0.0:5433`.

Fixes: https://github.com/yugabyte/yugabyte-db/issues/4726


ramkumarvs (approved):
LG

iSignal:
Thanks Ram for the review.

@baba230896 : maybe wait for @Arnav15's go ahead since this may require a corresponding change in the platform code. We might need to wait until the corresponding changes is ready to check this in.

iSignal:
@baba230896 : it looks like we are passing in insecure vs insecureClient to distinguish between the allow_insecure_conns flag to master vs tserver. Let's make this one single flag as before.

baba230896:
This is my understanding from the docs and I guess there is some difference between docs and reality.
I've tested the changes on the older version when I raised the PR but today I tested with v2.2.2.0 and got different results.

https://docs.yugabyte.com/latest/secure/tls-encryption/client-to-server/
https://docs.yugabyte.com/latest/secure/tls-encryption/server-to-server/

**Docs says:**

------
**For Master:** 

> 


allow_insecure_connections | YB-Master only | Set to false to disallow any service with unencrypted communication from joining this cluster. Default value is true. Note that this flag requires --use_node_to_node_encryption to be enabled.
-- | -- | --


**Derived:** The `allow_insecure_connections=false` flag prevents to join the insecure tservers to the clusters.

**For Tserver:** 

> 

--allow_insecure_connections | YB-TServer | Set to false to disallow any client with unencrypted communication from joining this cluster. Default value is true. Note that this flag requires --use_client_to_server_encryption to be enabled.
-- | -- | --

To enable access control, follow these steps, start the yb-tserver services with the following flag (described above):

`--use_client_to_server_encryption=true`

This flag enables both encrypted and unencrypted clients to connect to the cluster.

To prevent clients without the appropriate encryption from connecting, you must add the following flag:

`--allow_insecure_connections=false`



**Derived:** The `allow_insecure_connections=false` flag prevents to join the insecure client to tserver and using this `--use_client_to_server_encryption=true` flag can enable the connection for both types of clients.

------
I've implemented these two `insecure, insecureClient` to handle these two scenarios and `--use_client_to_server_encryption=true` for both the scenarios because as per the docs, it allows both types of traffics:

1. `insecure: false (master), insecureClient: true (tserver)` - All clients are allowed (With or Without certificates). In this, the cluster has encrypted communication but both types (insecure and secure) of clients can communicate to tserver.
2. `insecure: true (master), insecureClient: false (tserver)` - Clients with Certificates are only allowed. In this, both types (insecure and secure) of Tserver can join the cluster but the only secure client can connect to tserver.

But when I tested today (v2.2.2.0), if the tserver has `--use_client_to_server_encryption=true` then whether tserver has `--allow_insecure_connections=true|false`, it allows only secure clients to connect with tserver and if `--use_client_to_server_encryption=false` at tserver then only insecure can connect, so as result we can connect only one type of client at a time with tserver. But docs say something different from the above-mentioned result.

Will do some more thorough testing and let you know.

iSignal (changes requested):
Thanks for the details, Yatish.

Let's go back to making this a single flag (so no insecureClient option). There doesn't seem to be a good scenario to make the masters do insecure while the tservers do not or vice versa.

iSignal (approved):

PR #16, opened 2020-07-13
Add support for specifying security context

### LINKED ISSUE #15: Add support for specifying securitycontext for platform deployments
If we want to run platform as nonRoot user, we need to be able to specify the security context at both container level and overall context.

ramkumarvs:
Add support for specifying security context when bringing up platform 

tested by running
`helm install yugaware --name yw-ram --set=image.tag=2.1.8-ubi-b6 --set=image.repository=nchandrappa/yugaware --set=securityContext.enabled=true --wait`

By default it won't be enabled.

closes #15


stable/yugaware/values.yaml
```diff
@@ -29,6 +29,13 @@ tls:
...
 
+## yugaware pod Security Context
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+securityContext:
+  enabled: false
+  fsGroup: 1001
+  runAsUser: 1001
```
danielisen, line 37: @ramkumarvs will the yugabyte user always have UID of 1001?

nchandrappa: There is a further refinement that can be done by specifying `runAsUser: 'MustRunAsNonRoot'` based on the k8s pod-security-policy docs [here](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups). Looks like `MustRunAsNonRoot` need to be used in conjunction with `allowPrivilegeEscalation=false` attribute. This will force users to issue the pod with non-root numeric user-id or else the user-id used when creating the container will be used.

nchandrappa (approved):
Code changes look good to me. 

Although, there is a further refinement that can be done by specifying `runAsUser: 'MustRunAsNonRoot'` based on the k8s pod-security-policy docs [here](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups). Looks like `MustRunAsNonRoot` need to be used in conjunction with `allowPrivilegeEscalation=false` attribute. This will force users to issue the pod with non-root numeric user-id or else the user-id used when creating the container will be used.

stable/yugaware/values.yaml
ramkumarvs: @daniel-yb yes we explicitly set that in the RHEL docker image. again as in the helm chart, they can override any of these values.

PR #17, opened 2020-08-04
Add python3 decoding when generating kubeconfig

WesleyW:
When generating a kubeconfig through python3, the token is a byte which needs to be decoded or else the kube_config will not work properly.

### DIFF
--- a/stable/yugabyte/generate_kubeconfig.py
+++ b/stable/yugabyte/generate_kubeconfig.py
@@ -63,7 +63,7 @@ with tempfile.NamedTemporaryFile() as ca_crt_file:
                        '--certificate-authority={}'.format(ca_crt_file.name)]
     run_command(set_cluster_cmd, as_json=False)
 
-user_token = base64.b64decode(secret_data['data']['token'])
+user_token = base64.b64decode(secret_data['data']['token']).decode('utf-8')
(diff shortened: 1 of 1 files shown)


stable/yugabyte/generate_kubeconfig.py
```diff
@@ -63,7 +63,7 @@ def run_command(command_args, namespace=None, as_json=True):
                        '--certificate-authority={}'.format(ca_crt_file.name)]
     run_command(set_cluster_cmd, as_json=False)
 
-user_token = base64.b64decode(secret_data['data']['token'])
```
iSignal, line 66: we also do this at line 55. The best option would be to make run_command return strings all the time instead of bytes in python3 and strings in python3.

Maybe run it against python2 too to test that doing a decode('utf-8') returns the expected results.

WesleyW: @iSignal L55 actually does require a byte string since the default mode for `tempfile.NamedTemporaryFile()` is 'w+b'. I tested that it generates a file and eyeballed it afterwards on py2.

iSignal: oh ok, didn't realise that! I'd argue that converting it all out to string makes for less subtle code and is easier to understand but am ok with the current approach too.

iSignal (approved):

PR #25, opened 2020-09-21
Stop mounting thirdparty deps during helm deployment

danielisen:
Since we now rollup thirdparty deps in YW image, we shouldn't mount thirdparty volume from thirdparty-deps image during deployments


iSignal:
Looks good, can you add a test plan? Might be worth testing an upgrade scenario?

PR #28, opened 2020-09-24
Add tolerations to StatefulSet

### LINKED ISSUE #27: Tolerations is not added to chart
https://github.com/yugabyte/charts/blob/6e6ea5a2850ed8522fe3e4f1fb09cd25c15aecfb/stable/yugabyte/values.yaml#L181

This isn't actually present anywhere in the chart as far as I can tell.

twavv:
Fixes #27

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -187,6 +187,10 @@ spec:
       nodeSelector:
       {{ toYaml $root.Values.nodeSelector | indent 8 }}
       {{- end }}
+      {{- if $root.Values.tolerations }}
+      tolerations:
+{{ toYaml $root.Values.tolerations | indent 8 }}
+      {{- end }}
       affinity:
         # Set the anti-affinity selector scope to YB masters.
         {{ if $root.Values.AZ }}


stable/yugabyte/templates/service.yaml
```diff
@@ -186,6 +186,10 @@ spec:
       {{- if $root.Values.nodeSelector }}
       nodeSelector:
       {{ toYaml $root.Values.nodeSelector | indent 8 }}
+      {{- end }}
+      {{- if $root.Values.tolerations }}
+      tolerations:
+{{ toYaml $root.Values.tolerations | indent 8 }}
```
twavv, line 192: This has to be unindented to make the `| indent 8` work, apparently. Unsure if there's a better way to do it.

This is actually also kind of an issue for `nodeSelector`, but I think it would only show up if there are more than one label in the selector. If there's only one, the selector is visually overindented but still valid YAML:

```yaml
    spec:
      nodeSelector:
        # should be here...
              # but is actually here
              foo: bar
```

bhavin192: You are right, good catch :smile: . AFAIK better would be to have something like this,
```yaml
      {{- if $root.Values.tolerations }}
      tolerations:
      {{- toYaml $root.Values.tolerations | nindent 8 }}
      {{- end }}
```
The hyphen before `toYaml` will remove one line (along with our indentation). The `nindent` will add a new line and then indent our object by 8.

Also, as you mentioned the nodeSelector needs to be fixed.

bhavin192 (changes requested):
Can you also change the `tolerations: []` from values.yaml to `tolerations: {}` as it's not a list.

iSignal:
Along with the changes requested by @bhavin192 , could you please add a Test Plan to the commit summary with the helm command line used to test these changes? Thanks!

PR #31, opened 2020-09-30
Elevate hardcoded image repos/tags to values

maxwell-gregory:
I have moved image repository names and versions to values.yaml to be overridden by users that might want to use internal mirrors.

`thirdparty-deps` might be tricky if the repo is set to private

### DIFF
--- a/stable/yugaware/templates/service.yaml
+++ b/stable/yugaware/templates/service.yaml
@@ -100,7 +100,7 @@ spec:
             secretName: {{ .Release.Name }}-yugaware-tls-cert
         {{- end }}
       containers:
-        - image: postgres:11.5
+        - image: {{ .Values.postgres.repository }}:{{ .Values.postgres.tag }}
           name: postgres
           env:
             - name: POSTGRES_USER
@@ -128,7 +128,7 @@ spec:
               mountPath: /var/lib/postgresql/data
               subPath: postgres_data
         - name: prometheus
-          image: prom/prometheus:v2.2.1
+          image: {{ .Values.prometheus.repository }}:{{ .Values.prometheus.tag }}
           securityContext:
             runAsUser: 0
           volumeMounts:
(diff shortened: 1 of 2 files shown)


rkarthik007:
Hi @maxwell-gregory - could you please add yourself to [the contributors list](https://github.com/yugabyte/yugabyte-db/blob/master/CONTRIBUTORS.md) in case you haven't already done so? Thanks!

dotyjim-work:
This issue was discussed in https://github.com/yugabyte/yugabyte-db/issues/1634

iSignal:
Thanks for the contribution @maxwell-gregory !

1. Can you add a test plan to the summary of the PR, with the helm command line you used to test this?

2. It looks like the intention here would be primarily to customize the registry and not the image names, am I understanding this right? It would be good to split the customization into three parts:
* an image.registry value for each image (postgres, prometheus etc) that can be customized to the docker registry being used (docker.io, quay.io)
* the image names could also be customized, but I don't think that is a common requirement. If you think it is, please explain the use case.
* the image tag which also should be configurable for each image.

maxwell-gregory:
1. I ran a diff using a `helm template` as this would expose any template syntax or compilation issues. if you know of a better way please let me know
2. Your remarks are correct.  I do not think image names is a requirement as users dont want to package images for something they dont own

iSignal (changes requested):
Makes sense. Can you then change the config options to be just a repository (docker hub / quay) with defaults being docker hub / quay as they are no and a tag. So the repository would not include the image name?

For the Test Plan: doing a helm template sounds good. I would add the following text or similar to the PR description.

```
Test Plan:
1. Ran `helm template local/dir/.... > /tmp/new.yaml` and `helm template repo/dir/.... > /tmp/old.yaml` without my changes. `diff /tmp/old.yaml /tmp/new.yaml` shows no changes.
2. Ran the same with --set nginx.repository=quay.io and see the following diff
```

dotyjim-work:
@maxwell-gregory - Wanted to circle back on this PR.

First, thank you for posting this. We took the feedback to heart, but there were some growing pains on my side to getting this exact implementation shepherded in.

So it looks like the work got picked up in https://github.com/yugabyte/yugabyte-db/issues/1634 (we are thinking about how to coordinate this better, but most of the feature work is tracked in the issues in the main repo at this point in time), and was addressed in https://github.com/yugabyte/charts/commit/7e7857bc904cb9335bd4a175f872d9501c5e9d24 - via https://github.com/yugabyte/charts/pull/46.

Again, thank you for submitting this work, and I know that @rkarthik007 would love if you could you please add yourself to the [contributors list](https://github.com/yugabyte/yugabyte-db/blob/master/CONTRIBUTORS.md).

PR #34, opened 2020-10-01
WIP: Create Helm repository using GitHub pages

bhavin192:
-   Uses 'docs' directory as GitHub pages root.
-   Rename update_version.sh to update_chart_files.sh, this file takes care of updating versions in chart's source files.
-   update_version.sh does following things,
    -   If the released version is a latest one, it updates the source files from 'stable' directory and packages the new charts.
    -   If the released version is update to an old version,
        -   It tries to find if there is any tar for x.y.z. Update the tar if so. (Change in the w part of the version from x.y.z.w).
        -   Otherwise, it tries to find latest version from x.y series, unpacks the content and creates a new tar out of it by updating the version, container image tag etc. (Change in the z or w part of the version from x.y.z.w).
    -   Updates the Helm repository index when required.
-   Update the GitHub workflow to install Helm.


bhavin192:
We decided to not use this approach.

PR #35, opened 2020-10-05
Branch awareness

jharveysmith:
This creates separate branches for each release series (2.1, 2.2, etc).
Not sure how this will play with the helm repo site.  I'll have an alternative approaching using release directories in master branch shortly.

### DIFF
--- a/.github/workflows/update-version.yaml
+++ b/.github/workflows/update-version.yaml
@@ -3,6 +3,8 @@ on:
   repository_dispatch:
     types:
     - update-on-release
+env:
+  CURRENT_STABLE: 2.3
 jobs:
   update-version:
     if: github.event.client_payload.prerelease == 'false'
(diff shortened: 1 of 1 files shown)


.github/workflows/update-version.yaml
```diff
@@ -47,7 +71,7 @@ jobs:
         git add ./stable/yugabyte/Chart.yaml ./stable/yugaware/Chart.yaml \
                 ./stable/yugabyte/values.yaml ./stable/yugaware/values.yaml ./stable/yugabyte/app-readme.md
         git commit -m "Update the version to ${{steps.extract-version.outputs.yb_version}}"
-        git push origin ${{ github.ref }}
+        git push
```
svarnau, line 69: To be specific, maybe: git push origin HEAD

.github/workflows/update-version.yaml
```diff
@@ -3,12 +3,15 @@ on:
   repository_dispatch:
     types:
     - update-on-release
+env:
+  CURRENT_STABLE: 2.3
```
svarnau: As discussed, I think I prefer if the dispatcher side send parameter is_master, based on content of version.txt on master branch on that side.

.github/workflows/update-version.yaml
```diff
@@ -30,6 +33,27 @@ jobs:
         fi
         echo "Extracted the version number '${version_number}'."
         echo "::set-output name=yb_version::${version_number}"
+    - name: "Switch to correct branch"
+      id: extract-branch
+      run: |
+        rel=${{steps.extract-version.outputs.yb_version}}
+        a=( ${rel//./ } )
+        branch_name="${a[0]}.${a[1]}"
```
svarnau: Come to think of it, maybe we should do this calculation on dispatcher side and send a branch_name parameter? Then we won't need the is_master parameter, we just get teh branch name from yugabyte-db side.

svarnau (commented):
Looks good, but maybe we need to change yugabyte-db side first, to send new parameter.

.github/workflows/update-version.yaml
jharveysmith: If we do this, lets call it `yb_release` or maybe `yb_release_train` or `release_train` rather than branch.

svarnau (approved):

PR #37, opened 2020-10-09
[charts] Allow version update with specific docker tag.

svarnau:
Rather than lookup latest tag, allow command line argument.

May be used for local build as well as publishing via github action.

### DIFF
--- a/.ci/update_version.sh
+++ b/.ci/update_version.sh
@@ -10,13 +10,14 @@ function version_gt() {
 }
 
 
-if [[ $# -ne 1 ]]; then
-  echo "No arguments supplied. Please provide the release version" 1>&2
+if [[ $# < 1 || $# > 2 ]]; then
+  echo "Incorrect number of arguments. Usage: ${0%*/} <version> [<docker-tag>]" 1>&2
   echo "Terminating the script execution." 1>&2
   exit 1
 fi
 
 release_version="$1"
+docker_image_tag="$2"
 # appVersion mentioned in Charts.yaml
 current_version="$(grep -r "^appVersion" "stable/yugabyte/Chart.yaml" | awk '{ print $2 }')"
 if ! version_gt "${release_version}" "${current_version%-b*}" ; then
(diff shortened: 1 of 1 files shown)


.ci/update_version.sh
```diff
@@ -10,28 +10,31 @@ function version_gt() {
...
 
-# Find Docker image tag respective to YugabyteDB release version
-docker_image_tag_regex=[0-9]\+.[0-9]\+.[0-9]\+.[0-9]\+-b[0-9]\+
-docker_image_tag="$(python3 ".ci/find_docker_tag.py" "-r" "${release_version}")"
-if [[ "${docker_image_tag}" =~ ${docker_image_tag_regex} ]]; then
-  echo "Latest Docker image tag for '${release_version}': '${docker_image_tag}'."
-else
-  echo "Failed to parse the Docker image tag: '${docker_image_tag}'" 1>&2
-  exit 1
+if [[ -z "$docker_image_tag" ]]; then
```
jharveysmith, line 28: If `docker_image_tag` is always set (as we seem to force above with the check for two arguments), this entire section should be removed.

.ci/update_version.sh
```diff
@@ -10,28 +10,31 @@ function version_gt() {
...
-if [[ $# -ne 1 ]]; then
-  echo "No arguments supplied. Please provide the release version" 1>&2
+if [[ $# < 1 || $# > 2 ]]; then
+  echo "Incorrect number of arguments. Usage: ${0%*/} <version> [<docker-tag>]" 1>&2
   echo "Terminating the script execution." 1>&2
   exit 1
 fi
 
 release_version="$1"
+docker_image_tag="$2"
```
jharveysmith, line 20: Can you post an example of what gets passed to this script?

svarnau: Second arg can be empty. Above checks for 1 or 2 arguments. If only one is provided, $2 will be empty string.

svarnau: Sample usage:

.ci/update_version.sh 2.3.3.0 2.3.3.0-b22

svarnau: If we provide only one arg, we get old behavior of searching

jharveysmith (approved):

PR #40, opened 2020-11-02
Create a symlink to cores directory based on core_pattern

bhavin192:
Modifies the postStart hook to create a symlink to /mnt/disk0/cores at
the directory being used by kernel.core_pattern. This is useful when
the core_pattern's value points to some absolute path like
/var/lib/cores instead of a relative path and the path doesn't exist
inside the container.

Changing to bash as some of the bash constructs are being used. The
current container image's sh is just a symlink to bash.

Ref: https://github.com/yugabyte/yugabyte-db/issues/4385#issuecomment-717656848

### Scenarios tested
- Installed the chart on nodes having `kernel.core_pattern=/var/lib/test/cores123/core.%e.%p.%t`.
- Exec into the pod,
  ```console
  $ ls -ll /var/lib/test
  total 0
  lrwxrwxrwx 1 root root 16 Nov  2 13:45 cores123 -> /mnt/disk0/cores

  $ /home/yugabyte/dumpit 
  Hello World!
  Segmentation fault (core dumped)

  $ pwd
  /mnt/disk0/cores

  $ ls
  core.dumpit.78.1604324807
  ```

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -237,10 +237,19 @@ spec:
           postStart:
             exec:
               command:
-                - "sh"
+                - "bash"
                 - "-c"
                 - >
                   mkdir -p /mnt/disk0/cores;
+                  pattern="$(sysctl kernel.core_pattern --values || true)";
+                  core_dir="${pattern%/*}";
+                  if [ "${core_dir:0:1}" = "/" ]; then
+                    if [ ! -d "${core_dir}" ]; then
+                      # create the parent directory
+                      mkdir -p "${core_dir%/*}" || true;
+                      ln -s "/mnt/disk0/cores" "${core_dir}" || true;
+                    fi;
+                  fi;
                   mkdir -p /mnt/disk0/yb-data/scripts;
                   if [ ! -f /mnt/disk0/yb-data/scripts/log_cleanup.sh ]; then
                     if [ -f /home/yugabyte/bin/log_cleanup.sh ]; then


ajcaldera1 (approved):
LGTM.

stable/yugabyte/templates/service.yaml
```diff
@@ -218,10 +218,19 @@ spec:
...
                 - "-c"
                 - >
                   mkdir -p /mnt/disk0/cores;
+                  pattern="$(sysctl kernel.core_pattern --values)";
+                  core_dir="${pattern%/*}";
+                  if [ "${core_dir:0:1}" = "/" ]; then
+                    if [ ! -d "${core_dir}" ]; then
+                      # create the parent directory
+                      mkdir -p "${core_dir%/*}";
+                      ln -s "/mnt/disk0/cores" "${core_dir}";
```
iSignal: we should make sure none of these can cause the postStart to fail, because that would cause the container to never reach running state. Since this step is non-critical, I would add a `|| true` to it

iSignal (approved):

stable/yugabyte/templates/service.yaml
bhavin192: Yep, made the changes.

stable/yugabyte/templates/service.yaml
```diff
@@ -218,10 +218,19 @@ spec:
           postStart:
             exec:
               command:
-                - "sh"
+                - "bash"
                 - "-c"
                 - >
                   mkdir -p /mnt/disk0/cores;
+                  pattern="$(sysctl kernel.core_pattern --values || true)";
```
iSignal, line 244: if the sysctl fails, pattern is an empty string right, is that fine?

iSignal (approved):

stable/yugabyte/templates/service.yaml
bhavin192: Yes, that's fine. It won't go to the inside block where mkdir and ln is run.

PR #42, opened 2020-11-04
To configure username and password using Helm

baba230896:
## Summary

Added implementation to configure the username and password for YSQL and YCQL using Helm.

**Flow:**
Added post-install hook job to set up the credentials. The post-install hook job gets the script from ConfigMap. The hook job sets the custom credentials in the environment vars & the `setup credentials` script uses those environment vars for configuration.

[EDITED]
**Note:** 

- The changes work with TLS and Non-TLS clusters.

~~- The changes only work with the Non-TLS cluster.~~
~~- Compatible Changes for TLS is in-progress.~~

## Test Plan:
The changes have been tested with the mentioned env vars combination in `yugabyted` [reference docs](https://docs.yugabyte.com/latest/reference/configuration/yugabyted/#environment-variables). 

The following commands to test the changes (These changes are from the demo branch, and the branch is in development)-

1. The following commands are used to spin up the cluster with the custom `YSQL` password.
```sh
helm repo remove yugabytedb
helm repo add yugabytedb https://baba230896.github.io/charts
helm repo update
helm search repo yugabytedb/yugabyte
helm install yb-demo yugabytedb/yugabyte \
    --set resource.master.requests.cpu=150m,resource.master.requests.memory=100Mi,storage.master.size=3Gi \
    --set resource.tserver.requests.cpu=150m,resource.tserver.requests.memory=100Mi,storage.tserver.size=3Gi \
    --set replicas.tserver=3,replicas.master=3 \
    --set authCredentials.ysql.password=sqlpassword \
    --namespace ys-helm 
```
2. Verify the new password (Provide the new password in prompt).
```sh
kubectl exec -n ys-helm -it yb-tserver-0 -- ysqlsh -h yb-tserver-0.yb-tservers.ys-helm
```

## Fixes:
https://github.com/yugabyte/yugabyte-db/issues/5663

### DIFF
--- a/stable/yugabyte/templates/hooks/setup-credentials-job.yaml
+++ b/stable/yugabyte/templates/hooks/setup-credentials-job.yaml
@@ -1,0 +1,66 @@
+{{- if or .Values.authCredentials.ycql.user .Values.authCredentials.ycql.password .Values.authCredentials.ycql.keyspace .Values.authCredentials.ysql.password .Values.authCredentials.ysql.user .Values.authCredentials.ysql.database }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "yugabyte.fullname" . }}-setup-credentials
+  namespace: "{{ .Release.Namespace }}"
+  labels:
+    app: "setup-credentials"
+    release: {{ .Release.Name | quote }}
+    chart: "{{ .Chart.Name }}"
+    component: "{{ .Values.Component }}"
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  backoffLimit: 2
+  template:
+    metadata:
+      name: "setup-credentials"
+      labels:
+        app: "setup-credentials"
+        release: {{ .Release.Name | quote }}
+        chart: "{{ .Chart.Name }}"
+        component: "{{ .Values.Component }}"
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: setup-credentials
+        image: "{{ .Values.Image.repository }}:{{ .Values.Image.tag }}"
+        env:
+        {{- if .Values.authCredentials.ysql.user }}
+        - name: YSQL_USER
+          value: "{{ .Values.authCredentials.ysql.user }}"
+        {{- end }}
+        {{- if .Values.authCredentials.ysql.password }}
+        - name: YSQL_PASSWORD
+          value: "{{ .Values.authCredentials.ysql.password }}"
+        {{- end }}
+        {{- if .Values.authCredentials.ysql.database }}
+        - name: YSQL_DATABASE
+          value: "{{ .Values.authCredentials.ysql.database }}"
+        {{- end }}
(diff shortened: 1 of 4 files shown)


stable/yugabyte/templates/hooks/setup-credentials-job.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - name: YCQL_KEYSPACE
+          value: "{{ .Values.authCredentials.ycql.keyspace }}"
+        {{- end }}
+        {{- if .Values.tls.enabled }}
+        - name: SSL_CERTFILE
+          value: "/root/.yugabytedb/root.crt"
+        {{- end }}
+        command:
+        - 'bash'
+        - '/home/yugabyte/bin/setup-credentials/setup-credentials.sh'
```
iSignal, line 62: does the return value of this cmd matter?

stable/yugabyte/templates/setup-credentials-configmap.yaml
```diff
@@ -0,0 +1,259 @@
...
+        -c "\\conninfo"
+      )
+
+      if [[ "$4" == "true" ]]; then
+        ysql_cmd+=("sslmode=require")
+      fi
+
+      echo "${ysql_cmd[@]}"
+      while ! "${ysql_cmd[@]}"; do
+        sleep 5s
```
iSignal, line 91: would be good to log something like "Waiting for YSQL interface to become available". 

Was thinking that if the cluster never becomes healthy, this hook will never complete and the pod will never become Ready/Running. Will kubectl describe be able to tell us what is happening in that case?

baba230896: log is already added at line no. 217.

This part exists in the `waitUntilHealthy` function and, this function is used with a timeout of 3m at line no.218. The script will wait for 3 mins to get the cluster healthy, but somehow the cluster is unhealthy then the script exit with status 1. And the "backoffLimit" is 2 so K8s will execute the hook two times in case of failure and even after 2 executions hook failed then the `kubectl/helm` exit and return the error something like "post hook failed". We're removing the post-install hook job pod only on success so in case of failure, we can check the logs of the job and can do some debugging.

stable/yugabyte/templates/hooks/setup-credentials-job.yaml
baba230896: Yeah because of this only container will know the script is executed successfully or not.

iSignal: Makes sense - but it looks like when this cmd fails, the entire release is considered failed? Can you try that scenario to make sure that the error is easy to diagnose?

iSignal (approved):

stable/yugabyte/templates/hooks/setup-credentials-job.yaml
baba230896: In this case, the debugging is easy because, in the end, it gives the error something like this 'post-hook failed'. We're removing the post-install hook job's pod only on success so, in case of failure, we can check the logs of the job and can do further debugging.

PR #48, opened 2020-12-14
[yugaware] Changes for compatibility with OpenShift

### LINKED ISSUE #6431: [YW] Make YW service account optionally not require cluster role binding
https://github.com/yugabyte/charts/blob/master/stable/yugaware/templates/rbac.yaml has a list of ClusterRoleBindings. These need to be optionally turned off in order to allow YW to start up in an openshift project without cluster level role bindings.

### LINKED ISSUE #6548: [YW] [openshift] nginx image has issues in restricted scc
@bhavin192 identified two issues with the nginx config in the YW helm chart

1. It uses port 80
2. The image relies on cache dirs being accessible by the user the nginx process is run as. For (2), we could instead use https://hub.docker.com/r/nginxinc/nginx-unprivileged/tags?page=1&ordering=last_updated&name=1.17.4

### LINKED ISSUE #6617: [k8s] [yw] Move logs dir for YW to volume mount
For openshift, we also need to set ANSIBLE_LOCAL_TMP

bhavin192:
-   Use nginxinc/nginx-unprivileged image and use port 8443 and 8080 of the container. The ports exposed by the service remain the same.
-   Add a new option `rbac.enabled` which makes it possible to turn off the ClusterRole and ClusterRoleBinding creation.
-   Add a new option `ocpCompatibility.enabled` which removes runAsUser: 0 securityContext from prometheus container. Not changing the default securityContext here to make sure that old deployments continue to function with this change.
-   dnsmasq binds to 5353 instead of 53 now, remove the `--default-resolver` flag as well, otherwise dnsmasq modifies the `/etc/resolv.conf` and adds itself as nameserver as 127.0.0.1 without mentioning the port. This results in broken DNS setup inside the container.
    
    The /etc/resolv.conf looks like this when --default-resolver is specified
    
    ```conf
    nameserver 127.0.0.1 # added by go-dnsmasq
    # disabled by go-dnsmasq # nameserver 10.0.0.10
    search yw-test.svc.cluster.local svc.cluster.local cluster.local …
    options ndots:5
    ```

### Scenarios tested

-   Installed latest chart and created one universe in a GKE cluster.
    
    ```
    helm3 install yw yugabytedb/yugaware -n yw-test --set image.tag=2.5.0.0-b2 --set tls.enabled=true
    ```
-   Upgraded it with same image and new changes. Proxy, the Kubernetes service and other basic functionality is working.
    
    ```
    helm3 upgrade yw -n yw-test ~/src/yugabyte-charts/stable/yugaware/ 
    ```
-   Tested same steps with 2.5.0.0-b2 to latest build YW image (from source) and with TLS.

- Tested the same setup on OCP with 
   ```
   helm3 install yw-test ~/src/yugabyte-charts/stable/yugaware/ --set ocpCompatibility.enabled=true --set rbac.create=false
   ```

Fixes https://github.com/yugabyte/yugabyte-db/issues/6548 Fixes https://github.com/yugabyte/yugabyte-db/issues/6431 Fixes https://github.com/yugabyte/yugabyte-db/issues/6617

### DIFF
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -32,6 +32,8 @@ data:
     pidfile.path = "/dev/null"
     play.evolutions.enabled=false
     play.modules.enabled += "org.flywaydb.play.PlayModule"
+    play.logger.includeConfigProperties=true
+    log.override.path = "/opt/yugaware_data/logs"
 
     db {
       default.url="jdbc:postgresql://{{ eq .Values.ip_version_support "v6_only" | ternary "[::1]" "127.0.0.1" }}:5432/"${POSTGRES_DB}
@@ -122,12 +124,12 @@ data:
   default.conf: |
     server {
 {{- if .Values.tls.enabled }}
-      listen       443 ssl;
+      listen       8443 ssl;
       ssl_certificate /opt/certs/server.crt;
       ssl_certificate_key /opt/certs/server.key;
       server_name  {{ .Values.tls.hostname }};
 {{- else }}
-      listen       {{ eq .Values.ip_version_support "v6_only" | ternary "[::]:80" "80" }};
+      listen       {{ eq .Values.ip_version_support "v6_only" | ternary "[::]:8080" "8080" }};
       server_name  {{ .Values.tls.hostname }};
 {{- end }}
       proxy_http_version 1.1;
@@ -150,7 +152,7 @@ data:
       }
 
       location ~ "^/proxy/(.*.svc.cluster.local):([0-9]{4,6})/(.*)$" {
-        resolver {{ eq .Values.ip_version_support "v6_only" | ternary "[::1]:53" "127.0.0.1:53 ipv6=off"  }};
+        resolver {{ eq .Values.ip_version_support "v6_only" | ternary "[::1]:53" "127.0.0.1:5353 ipv6=off"  }};
         proxy_pass "http://$1:$2/$3$is_args$args";
         sub_filter "http://" "/proxy/";
         sub_filter "href='/" "href='/proxy/$1:$2/";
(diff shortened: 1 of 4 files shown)


stable/yugaware/values.yaml
```diff
@@ -30,8 +30,8 @@ image:
 
   nginx:
     registry: ""
-    tag: 1.17.4
-    name: nginx
+    tag: 1.17.4-amd64
+    name: nginxinc/nginx-unprivileged
```
iSignal, line 34: minor: can you comment if the images work the exact same way (entrypoint, config) other than the fact that it is using non-root-owned dirs internally? so if an existing non-openshift user is using a copy of the nginx image from their local enterprise repo, they could continue to use it by setting it to the previous values and nothing would break.

stable/yugaware/templates/configs.yaml
```diff
@@ -32,6 +32,8 @@ data:
     pidfile.path = "/dev/null"
     play.evolutions.enabled=false
     play.modules.enabled += "org.flywaydb.play.PlayModule"
+    # play.logger.includeConfigProperties=true
+    # log.override.path = "/opt/yugaware_data/logs"
```
iSignal: do you want to still keep these commented? setting java properties seems to be more involved, right?

iSignal (approved):

stable/yugaware/values.yaml
bhavin192: Yes, that is correct. They can continue to use the old image. I cross checked this by setting the value.

stable/yugaware/templates/configs.yaml
bhavin192: If we uncomment these, then we will see blank `/logs` page. Basically we need, https://github.com/yugabyte/yugabyte-db/pull/6687

iSignal: Makes sense. Now that the other change is in, let's uncomment and merge this PR too.

PR #50, opened 2020-12-16
Modified to deploy multiple YB clusters in the single namespace

### LINKED ISSUE #6355: [k8s] Support deploying multiple helm yb charts in single namespace

baba230896:
### Summary

- Modified the Helm charts for multiple YB deployments in the single namespace.
- Now Helm Charts have two naming conventions, the old naming convention will be used by default, and to tweak the naming convention for the deployment of multiple YB clusters in the single namespace use `oldNamingStyle: false` or  `--set oldNamingStyle=false`.
- Migration won't possible for the old naming convention to the new naming convention due to `volumeClaimTemplates` because it creates the PVC every time whenever it gets a new combination of `given name + pod name`.

### Test Plan:

The following test scenarios have been tested so far.
**Test Environment:** OpenShift and GKE, `ysqlsh` with Northwind DB, and tested externally through EC2 instance.

1. Created a YB cluster with the old naming convention, upgraded with the `quay.io/yugabyte/yugabyte-itest:2.5.1.0-b90` image, and at last, upgraded with TLS flag. 
2. Created 2 YB clusters with the new naming convention, upgraded with the `quay.io/yugabyte/yugabyte-itest:2.5.1.0-b90` image, and at last, upgraded with TLS flag. 
3. Created a YB cluster with the new naming convention, upgraded with the `quay.io/yugabyte/yugabyte-itest:2.5.1.0-b90` image, and then upgraded with TLS flag but with the wrong path of certs to test rollback. And at last, after rollback, again upgraded with TLS flag.
4. Created a YB cluster using charts hosted on `charts.yugabyte.com`, upgraded with the `quay.io/yugabyte/yugabyte-itest:2.5.1.0-b90` image using these changes (keeping `oldNamingStyle: true`), and at last, upgraded with TLS flag. 
5. Created a YB cluster using charts hosted on `charts.yugabyte.com`, and try to upgrade without any change using these changes (keeping `oldNamingStyle: true`). And as expected, the pods haven't restarted.
6. Created a YB cluster with ServiceMonitor using charts hosted on `charts.yugabyte.com`, upgraded with the `quay.io/yugabyte/yugabyte-itest:2.5.1.0-b90` image using these changes (keeping `oldNamingStyle: true`), and at last, upgraded with TLS flag. 
7. Created a YB cluster with ServiceMonitor (With 53 Characters Release name) using charts hosted on `charts.yugabyte.com`, upgraded with the `quay.io/yugabyte/yugabyte-itest:2.5.1.0-b90` image using these changes (keeping `oldNamingStyle: true`), so the changes will remove the old SM and deploy the new SM using the new naming convention because previously `_helpers.tpl` trimming the 63 characters but now it is 43 characters. And throughout the upgrading process pod won't restart because we aren't upgrading the naming convention for pods.
8. Created a YB cluster with ServiceMonitor (new naming convention), upgraded with the `quay.io/yugabyte/yugabyte-itest:2.5.1.0-b90` image, and at last, upgraded with TLS flag. 

fixes https://github.com/yugabyte/yugabyte-db/issues/6355

### DIFF
--- a/stable/yugabyte/templates/_helpers.tpl
+++ b/stable/yugabyte/templates/_helpers.tpl
@@ -3,26 +3,64 @@
 Expand the name of the chart.
 */}}
 {{- define "yugabyte.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 43 | trimSuffix "-" -}}
 {{- end -}}
 
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
+The components in this chart create additional resources that expand the longest created name strings.
+The longest name that gets created of 20 characters, so truncation should be 63-20=43.
 */}}
 {{- define "yugabyte.fullname" -}}
 {{- if .Values.fullnameOverride -}}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- .Values.fullnameOverride | trunc 43 | trimSuffix "-" -}}
 {{- else -}}
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- if contains $name .Release.Name -}}
-{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- .Release.Name | trunc 43 | trimSuffix "-" -}}
 {{- else -}}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 43 | trimSuffix "-" -}}
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/* Generate common labels */}}
+{{- define "yugabyte.labels" }}
+heritage: {{ .Values.helm2Legacy | ternary "Tiller" (.Release.Service | quote) }}
+release: {{ .Release.Name | quote }}
+chart: {{ .Values.oldNamingStyle | ternary .Chart.Name (include "yugabyte.chart" .) | quote }}
+component: {{ .Values.Component | quote }}
+{{- if .Values.commonLabels}}
+{{ toYaml .Values.commonLabels }}
+{{- end }}
+{{- end }}
+
+{{/* Generate app label */}}
+{{- define "yugabyte.applabel" }}
+{{- if .root.Values.oldNamingStyle }}
+app: "{{ .label }}"
+{{- else }}
+app.kubernetes.io/name: "{{ .label }}"
+{{- end }}
+{{- end }}
+
+{{/* Generate app selector */}}
+{{- define "yugabyte.appselector" }}
+{{- if .root.Values.oldNamingStyle }}
+app: "{{ .label }}"
+{{- else }}
+app.kubernetes.io/name: "{{ .label }}"
+release: {{ .root.Release.Name | quote }}
+{{- end }}
+{{- end }}
+
+{{/* Create Volume name */}}
+{{- define "yugabyte.volume_name" -}}
+{{- printf "%s-datadir" (include "yugabyte.fullname" .) -}}
+{{- end -}}
+
 {{/*
 Derive the memory hard limit for each POD based on the memory limit.
 Since the memory is represented in <x>GBi, we use this function to convert that into bytes.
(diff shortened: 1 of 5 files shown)


stable/yugabyte/templates/service.yaml
```diff
@@ -205,10 +187,21 @@ spec:
...
                   - "{{ .label }}"
+                {{- else }}
+                - key: app.kubernetes.io/name
+                  operator: In
+                  values:
+                  - "{{ .label }}"
+                - key: release
+                  operator: In
+                  values:
+                  - {{ $root.Release.Name | quote }}
```
bhavin192, line 203: Having the release label here means that two YugabyteDB pods can be on same node if they are part of different releases. _Also some pod with old naming scheme can be with a pod of new naming scheme._

stable/yugabyte/templates/_helpers.tpl
```diff
@@ -3,26 +3,64 @@
 Expand the name of the chart.
 */}}
 {{- define "yugabyte.name" -}}
```
jvigil-yugabyte, line 5: Do we use this template anywhere? It looks like we might be able to delete it.

stable/yugabyte/templates/_helpers.tpl
```diff
@@ -3,26 +3,64 @@
...
+{{ toYaml .Values.commonLabels }}
+{{- end }}
+{{- end }}
+
+{{/* Generate app label */}}
+{{- define "yugabyte.applabel" }}
+{{- if .root.Values.oldNamingStyle }}
+app: "{{ .label }}"
+{{- else }}
+app.kubernetes.io/name: "{{ .label }}"
```
jvigil-yugabyte, line 45: Is there any reason why we wouldn't want to switch to using the recommended app name label all the time? ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

CC: @Arnav15, @iSignal

jvigil-yugabyte (changes requested):
In general, it would be an awesome feature to be able to support multiple YB clusters in a single namespace. This is a great step in that direction. However, it seems like this PR makes some bigger changes than just adding that capability. It modifies the naming convention for a few different objects (for example, `*-tls-client-cert` -> `*-client-tls`). In order to support the naming convention changes, it's my opinion that the extensive use of ternary operators add a lot of complexity to our chart overall and might not be worth it.

It seems like we could add support for running multiple YB clusters in the same namespace in a simpler way by creating an optional "namingPrefix" field in the values.yml file. If specified, the prefix will be prepended to all of the objects that need unique naming. Otherwise, an empty string will be prepended and it will have no effect. What do you think about that plan? Or, maybe I am missing some context on why these larger naming convention changes are necessary.

Also, maybe we want to split the template functions for generating the labels into a separate commit / PR? It's a nice refactor that cleans up our duplicate label generation everywhere and probably deserves to be it's own separate thing :100:

iSignal:
@jvigil-yugabyte : not sure I follow the line of thought. In general, there is already a unique naming prefix available to us - the release name, why would we introduce another? 

As for the complexity of choosing between old and new formats, it arises from the problem of maintaining compatibility with older versions, not from the approach of the PR itself. If we can introduce a breaking major version upgrade at some point, it allows us to get rid of the complexity but until then, we will have to support both formats.

jvigil-yugabyte:
@iSignal good point, I did not realize we had the release name as a unique prefix. We could definitely use that. My proposal was around simplifying the changes (for this PR / feature) by utilizing a unique prefix, rather than creating an entirely new naming convention.

For example, we could replace the changes like:
```
name: {{ $root.Values.oldNamingStyle | ternary "yugabyte-tls-client-cert" (printf "%s-client-tls" (include "yugabyte.fullname" $root)) }}
```
... with something like:
```
name: {{ printf "%syugabyte-tls-client-cert" .Values.namingPrefix }}
```

This way, we could retain the old naming convention, yet also add support for unique names (if so desired). Though, if we do want to eventually change our overall naming convention, then I agree, we will have to find some way to make a breaking change / upgrade to the helm chart.

stable/yugabyte/templates/service.yaml
```diff
@@ -263,9 +256,17 @@ spec:
           {{- if $root.Values.istioCompatibility.enabled }}
           - "--rpc_bind_addresses=0.0.0.0:7100"
           {{- else }}
+          {{- if $root.Values.oldNamingStyle }}
           - "--rpc_bind_addresses={{ printf "$(HOSTNAME).%s.$(NAMESPACE).svc.%s" .name $root.Values.domainName }}"
+          {{- else }}
+          - "--rpc_bind_addresses={{ printf "$(HOSTNAME).%s-%s.$(NAMESPACE).svc.%s" (include "yugabyte.fullname" $root) .name $root.Values.domainName }}"
```
iSignal, line 262: this pattern appears a bunch of times so it would be better to make this a utility function in helpers

iSignal (approved):

jvigil-yugabyte (approved):

jvigil-yugabyte:
After chatting w/ @iSignal, it seems we need to support backwards compatibility, so the `oldNamingStyle` flag is necessary.

PR #53, opened 2021-01-21
[YW] Removed nginx proxy_pass rules 

### LINKED ISSUE #6813: [yw] [k8s] Remove nginx proxy_pass rules in helm chart
We shouldn't need these nginx proxy_pass rules with the latest changes to have the proxy inside YW code itself.

baba230896:
### Summary:
Removed the Nginx proxy pass rules from `config.yaml` and dnsmasq from STS.

### Test Plan:
**Test condition:** Using `<YW-Service>/proxy/<Master:7000>/` should not open the YB master UI in the private browser window.

1. Deploy the YW from current helm charts (Branch: master) and created a YB Cluster. The test is failing then upgraded the YW with the changes and the test is successfully passed.
2. Deploy a YW from the changes and created a YB cluster. And the test passed.

fixes https://github.com/yugabyte/yugabyte-db/issues/6813

### DIFF
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -151,17 +151,6 @@ data:
       location / {
         proxy_pass http://{{ eq .Values.ip_version_support "v6_only" | ternary "[::1]" "127.0.0.1" }}:9000;
       }
-
-      location ~ "^/proxy/(.*.svc.cluster.local):([0-9]{4,6})/(.*)$" {
-        resolver {{ eq .Values.ip_version_support "v6_only" | ternary "[::1]:53" "127.0.0.1:5353 ipv6=off"  }};
-        proxy_pass "http://$1:$2/$3$is_args$args";
-        sub_filter "http://" "/proxy/";
-        sub_filter "href='/" "href='/proxy/$1:$2/";
-        sub_filter "href=\"/" "href=\"/proxy/$1:$2/";
-        sub_filter "src='/" "src='/proxy/$1:$2/";
-        sub_filter "src=\"/" "src=\"/proxy/$1:$2/";
-        sub_filter_once off;
-      }
     }
 ---
 apiVersion: v1
--- a/stable/yugaware/templates/service.yaml
+++ b/stable/yugaware/templates/service.yaml
@@ -227,10 +227,3 @@ spec:
             mountPath: /opt/certs/
             readOnly: true
           {{- end }}
-        - name: dnsmasq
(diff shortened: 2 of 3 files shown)


stable/yugaware/templates/service.yaml
```diff
@@ -227,10 +227,3 @@ spec:
             mountPath: /opt/certs/
             readOnly: true
           {{- end }}
-        - name: dnsmasq
-          image: {{ include "full_image" (dict "containerName" "dnsmasq" "root" .) }}
```
iSignal, line 231: there is an entry corresponding to this in values.yaml which also needs to be removed.

iSignal (commented):
Like the test to make sure that the proxy url cannot be loaded without logging in to YW. Can you confirm that it does work properly when logged in to YW after these changes?

stable/yugaware/templates/configs.yaml
```diff
@@ -151,17 +151,6 @@ data:
       location / {
         proxy_pass http://{{ eq .Values.ip_version_support "v6_only" | ternary "[::1]" "127.0.0.1" }}:9000;
       }
-
-      location ~ "^/proxy/(.*.svc.cluster.local):([0-9]{4,6})/(.*)$" {
-        resolver {{ eq .Values.ip_version_support "v6_only" | ternary "[::1]:53" "127.0.0.1:5353 ipv6=off"  }};
```
danielisen, line 156: We need to test the new internal YW proxy logic with IPv6 (I'm not worried about the format of the proxy request since it used DNS names, but I am not sure about the resolver). @iSignal thoughts (since you know more about this than I do)?

danielisen:
> Like the test to make sure that the proxy url cannot be loaded without logging in to YW. Can you confirm that it does work properly when logged in to YW after these changes?

I could be wrong but I think @baba230896 used a private browsing window to avoid a cached response, not to test authentication. Although the new proxy is authenticated, `<YW-Service>/proxy/<Master:7000>/` isn't the correct format anymore. Now it is `<YW-Service>/universes/<universeUUID>/proxy/<Master:7000>/`.

stable/yugaware/templates/configs.yaml
iSignal: This change should be independent of possible IPv6 resolver issues because it is essentially removing the unused /proxy link. For IPv6, we can get feedback from a run in an IPv6 k8s cluster to isolate possible issues but that shouldn't affect this diff.

iSignal:
confirmed that the new univ proxy URL was tested to work fine.

iSignal (approved):

PR #59, opened 2021-03-04
[Platform] Use OpenShift's Prometheus for container related metrics

### LINKED ISSUE #6611: [YW] [openshift] Metrics in container section do not show up in openshift YW
When YW is installed without cluster level role binding into an openshift project/namespace, YW prometheus fails to scrape metrics from system pods like kube state metrics and cadvisor metrics. 

https://github.com/yugabyte/charts/blob/597e05078cc32dde0bed828036a4e3f5454f1f9b/stable/yugaware/templates/configs.yaml#L240

bhavin192:
### [Platform] Use OpenShift's Prometheus for container related metrics

In OpenShift, there is already a Prometheus in the
openshift-monitoring namespace which collects data from:

- All the kubelet and their cadvisor endpoints.
- kube-state-metrics from same namespace.

To be able to scrape this Prometheus using its /federate endpoint we
need the cluster-monitoring-view ClusterRole. ref:
https://docs.openshift.com/container-platform/4.5/monitoring/cluster_monitoring/prometheus-alertmanager-and-grafana.html

- With this change we are fetching the only metrics which we need from
  the OCP's Prometheus.
- Remove the kubernetes-pods job as we don't use it, and it can make
  our Prometheus scrape unrelated pods from the cluster (the ones
  which have prometheus.io/scrape label). This can result in more disk
  utilization.
- Add pod_name and container_name labels. These have take of pod and
  container labels respectively. Kubernetes 1.16 has only pod and
  container
  labels. https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#removed-metrics
- Not using the kubernetes.io/cluster-service and addonmanager.*
  labels for the ClusterRoleBinding as those are not required in our
  case. ref: https://git.io/Jqvjn

Scenarios tested:
- Created a new universe in OpenShift, it is showing the container
  related graphs correctly in the UI.

Fixes yugabyte/yugabyte-db#6611

### [Platform] Add old mountpoints for backward compatibility

4ceb0ce changed few default mount points, and added command to create
symlinks at old paths. This doesn't work in case of OpenShift, as we
don't have access to create new directories in /opt. This can happen
in any Kubernetes deployment which runs the pods as non root.

```
ln: failed to create symbolic link '/opt/yugaware_data': Permission denied
```

To solve this, mounting the same volume subPath at both old as well as
new locations. And removing the ln command.

Scenarios tested:
- Upgraded an existing platform Helm release with these changes,
  universe creation is working as expected.

### DIFF
--- a/stable/yugaware/templates/rbac.yaml
+++ b/stable/yugaware/templates/rbac.yaml
@@ -7,6 +7,23 @@ metadata:
     kubernetes.io/cluster-service: "true"
     addonmanager.kubernetes.io/mode: Reconcile
 {{- if .Values.rbac.create }}
+{{- if .Values.ocpCompatibility.enabled }}
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Release.Name }}-cluster-monitoring-view
+  labels:
+    app: yugaware
+subjects:
+- kind: ServiceAccount
+  name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: cluster-monitoring-view
+  apiGroup: rbac.authorization.k8s.io
+{{- else }}
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
@@ -62,4 +79,5 @@ roleRef:
   kind: ClusterRole
   name: {{ .Release.Name }}
   apiGroup: rbac.authorization.k8s.io
+{{- end }}
 {{- end }}
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -175,6 +175,50 @@ data:
         scrape_interval:     10s
         evaluation_interval: 10s
     scrape_configs:
+      {{- if .Values.ocpCompatibility.enabled }}
+      - job_name: "ocp-prometheus-federated"
+        scheme: https
+
+        tls_config:
+          ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+          insecure_skip_verify: true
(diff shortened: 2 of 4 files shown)


bhavin192:
Hey @daniel-yb can you please take a look at the second change related to volumes: https://github.com/yugabyte/charts/pull/59/commits/cc605fd9395103014f893cc9dec79f6a1eaa51d4

stable/yugaware/templates/rbac.yaml
```diff
@@ -7,6 +7,23 @@ metadata:
     kubernetes.io/cluster-service: "true"
     addonmanager.kubernetes.io/mode: Reconcile
 {{- if .Values.rbac.create }}
+{{- if .Values.ocpCompatibility.enabled }}
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Release.Name }}-cluster-monitoring-view
```
iSignal, line 15: probably not a good idea to include this because it will fail for users with the default permissions right? We could add a subflag under ocpCompatibility like 

```
ocpCompabiliity:
   enabled: true
   metrics_enabled: true
```

bhavin192: The initial thought was, in those cases they can just set 
```
rbac:
  create: false
```

danielisen:
@bhavin192 changing to mounting the volume twice to old/new mount path instead of symlinking looks okay to me given the non-root permission issues. Although please rebase your diff on master to pick up my latest commit 803228a which also changes some stuff here too (hence your merge conflicts). I just want to see this diff rebased on master before giving it my stamp to make sure there aren't any unexpected changes between now and then.

iSignal (approved):

bhavin192:
Keeping the cp command as it is for now. We might need to consider using something like initContainer to reliably make this file available before any of Prometheus or platform container starts.

PR #63, opened 2021-04-21
[YW] Added initContainer to copy the prometheus configuration

### LINKED ISSUE #7515: Add init container to copy prometheus config config map -> persistent volume
Instead of copying the default prometheus config to YW storage during YW startup, we should be using an init container to do so such that we can ensure the file has been copied before YW/prometheus containers startup.

baba230896:
### Summary

- Added `initContainer` in the YW helm chart to copy the Prometheus configuration.

### Test Plan

- Deployed the YW & spin up the YB universe on GCP. Updated the Prometheus configuration & is reflected in the Prometheus.

fixes: https://github.com/yugabyte/yugabyte-db/issues/7515

### DIFF
--- a/stable/yugaware/templates/statefulset.yaml
+++ b/stable/yugaware/templates/statefulset.yaml
@@ -67,9 +67,7 @@ spec:
       initContainers:
         - image: {{ include "full_yugaware_image" . }}
           name: prometheus-configuration
-          command: [ "/bin/sh", "-c"]
-          args:
-            - "cp /default_prometheus_config/prometheus.yml /prometheus_configs/prometheus.yml"
+          command: ["cp", "/default_prometheus_config/prometheus.yml", "/prometheus_configs/prometheus.yml"]
           {{- if .Values.securityContext.enabled }}
           securityContext:
             runAsUser: {{ .Values.securityContext.runAsUser }}
@@ -194,8 +192,6 @@ spec:
           - name: yugaware-storage
             mountPath: /opt/yugabyte/prometheus/rules
             subPath: swamper_rules
-          - name: prometheus-config
-            mountPath: /default_prometheus_config
           - name: yugaware-storage
             mountPath: /prometheus_configs
             subPath: prometheus.yml
--- a/stable/yugaware/templates/statefulset.yaml
+++ b/stable/yugaware/templates/statefulset.yaml
@@ -64,6 +64,22 @@ spec:
           secret:
             secretName: {{ .Release.Name }}-yugaware-tls-cert
         {{- end }}
+      initContainers:
+        - image: {{ include "full_yugaware_image" . }}
+          name: prometheus-configuration
+          command: [ "/bin/sh", "-c"]
+          args:
+            - "cp /default_prometheus_config/prometheus.yml /prometheus_configs/prometheus.yml"
+          {{- if .Values.securityContext.enabled }}
+          securityContext:
+            runAsUser: {{ .Values.securityContext.runAsUser }}
+          {{- end }}
+          volumeMounts:
+          - name: prometheus-config
+            mountPath: /default_prometheus_config
+          - name: yugaware-storage
+            mountPath: /prometheus_configs
+            subPath: prometheus.yml
       containers:
         - image: {{ include "full_image" (dict "containerName" "postgres" "root" .) }}
           name: postgres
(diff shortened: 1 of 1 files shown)


stable/yugaware/templates/service.yaml
```diff
@@ -99,6 +99,22 @@ spec:
           secret:
             secretName: {{ .Release.Name }}-yugaware-tls-cert
         {{- end }}
+      initContainers:
+        - image: busybox:1.33.0
```
jvigil-yugabyte: In the Yugabyte helm chart, we require busybox:1.31 (https://github.com/yugabyte/charts/blob/master/stable/yugabyte/templates/service.yaml#L392). Requiring two separate busybox images is a bit strange, especially for airgapped installations which will require the administrator to provide both. Should we use the same image in both places? Also, would we want to make this image + tag configurable via the values.yaml file like our other images? CC: @iSignal

jvigil-yugabyte: Also, just noticed https://github.com/yugabyte/charts/pull/54, which removes the other busybox image and replaces it with the yugabyte image. Maybe we would we want to use the yugaware image here, then?

iSignal: Agree that reusing an existing container would make it easier to deploy in offline environments where specific approved containers from an internal repo can be used.

iSignal (changes requested):
use a diff base image

baba230896:
@iSignal 
Could you please review it once whenever you've time? This is a very straightforward change.

stable/yugaware/templates/statefulset.yaml
```diff
@@ -64,6 +64,22 @@ spec:
           secret:
             secretName: {{ .Release.Name }}-yugaware-tls-cert
         {{- end }}
+      initContainers:
```
skorobogatydmitry, line 67: Suggestion: https://github.com/elastic/helm-charts/blob/master/logstash/templates/statefulset.yaml#L149

This approach is more flexible.

stable/yugaware/templates/statefulset.yaml
```diff
@@ -64,6 +64,22 @@ spec:
...
+        - image: {{ include "full_yugaware_image" . }}
+          name: prometheus-configuration
+          command: [ "/bin/sh", "-c"]
+          args:
+            - "cp /default_prometheus_config/prometheus.yml /prometheus_configs/prometheus.yml"
```
skorobogatydmitry: Why not use `cp` directly ?

stable/yugaware/templates/statefulset.yaml
```diff
@@ -130,7 +146,7 @@ spec:
           {{- end }}
           command: [ "/bin/bash", "-c"]
           args:
-            - "cp /default_prometheus_config/prometheus.yml /prometheus_configs/prometheus.yml && bin/yugaware -Dconfig.file=/data/application.docker.conf"
```
skorobogatydmitry, line 133: So we don't need [this](https://github.com/yugabyte/charts/pull/63/files#diff-58c816a5badea1634d1f7fd6bc835152552ed86e143de953e41b0953a9313deeR197-R198) anymore, right ?

stable/yugaware/templates/statefulset.yaml
iSignal: is the suggestion that we customize the init containers through values? For our current purpose, the init container here seems reasonable.

bhavin192: Yes, I agree with Sanketh. The current initContainer is something which is always going to be there in every chart installation.

stable/yugaware/templates/statefulset.yaml
baba230896: Yes, we can use `cp` directly. Will update it.

stable/yugaware/templates/statefulset.yaml
baba230896: I guess this is something I missed while rebasing to remove the merge conflicts. Indeed, I'll remove the `volumeMount` related to default Prometheus conf in YW STS Pod Spec.

baba230896:
@skorobogatydmitry Could you please review it again?

stable/yugaware/templates/statefulset.yaml
skorobogatydmitry: Yes, suggestion was to have `initContainers` in Values.
This allows to tune pods without re-building image and overall helpful.
And you're right - it's not required for this particular patch.

skorobogatydmitry:
> @skorobogatydmitry Could you please review it again?

Sure!
There is "request review" button in `Reviewers` section - this works better to let reviewer know to re-visit PR.

skorobogatydmitry (approved):
LGTM! Please collect one more approval.

bhavin192 (approved):
This looks good to me!

PR #66, opened 2021-05-07
[platform] Restart the pod if the configuration changes

### LINKED ISSUE #8240: [Platform] Updating certain values.yaml entries for Platform helm charts does not trigger a pod restart
Repro steps:
1. Deploy helm chart for platform as documented at https://docs.yugabyte.com/latest/yugabyte-platform/install-yugabyte-platform/install-software/kubernetes/
2. Update values.yaml entries like helm.timeout and others at https://github.com/yugabyte/charts/blob/a844371c3a54ca997c2712bd26cf91b7780fb336/stable/yugaware/templates/configs.yaml#L74. Run a `helm upgrade` after changing these values
3. Notice that config does not take effect, YW pod is not restarted.

bhavin192:
-   This adds checksum of the configs.yaml file to pod metadata as annotation. This causes it to change if there are any changes in the configuration, and results in a restart of the pod. https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
-   Moving the global-config outside from configs.yaml, as it has a random string in it, which causes a restart everytime as the checksum keeps changing.

Test plan:

-   Deployed platform chart in a Kubernetes cluster without this change.
-   Upgraded the Helm release with this change (the pod restarts as expected).
-   Add --set helm.timeout=1000 during the helm upgrade (the pod restarts as expected)
-   Did another upgrade with same command as above (the pod does not restart as the configuration is the same).

Fixes https://github.com/yugabyte/yugabyte-db/issues/8240

### DIFF
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -1,19 +1,4 @@
 # Copyright (c) YugaByte, Inc.
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: {{ .Release.Name }}-yugaware-global-config
-  labels:
-    app: {{ template "yugaware.name" . }}
-    chart: {{ template "yugaware.chart" . }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Values.helm2Legacy | ternary "Tiller" (.Release.Service | quote) }}
-data:
-  postgres_user: "postgres"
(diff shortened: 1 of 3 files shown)


iSignal (approved):
This is a nice trick!

iSignal:
@bhavin192 : can we backport this to the 2.6 branch of charts too?

PR #71, opened 2021-06-06
Add optional argument to allow setting pull secret file

skorobogatydmitry:


stable/yugaware/values.yaml
```diff
@@ -11,7 +10,10 @@ image:
   repository: quay.io/yugabyte/yugaware
   tag: 2.5.1.0-b153
   pullPolicy: IfNotPresent
-  pullSecret: yugabyte-k8s-pull-secret
+  pullSecret: yugabyte-k8s-pull-secret-1
```
bhavin192: We should not change the default name here. If someone has existing deployment, and they upgrade, image pull will fail in this case.

skorobogatydmitry: Good catch!

skorobogatydmitry:
Any other comments?
@bhavin192 @iSignal @rtsisyk

stable/yugaware/values.yaml
```diff
@@ -12,6 +11,9 @@ image:
   tag: 2.5.1.0-b153
   pullPolicy: IfNotPresent
   pullSecret: yugabyte-k8s-pull-secret
+  # Docker config JSON File name
+  # If set, this file content will be used to automatically create secred named as above
```
iSignal: s/secred/secret

iSignal (approved):
Looks good to me. @bhavin192 : please request changes if you have any other comments?

stable/yugaware/values.yaml
```diff
@@ -12,6 +11,9 @@ image:
...
+  # pullSecretFile:
```
```suggestion
  ## Docker config JSON File name
  ## If set, this file content will be used to automatically create secret named as above
  # pullSecretFile:
```

Just the convention of having comments with double hash and actual YAML in single hash.

bhavin192 (approved):
Only a minor formatting change. Else looks good.

PR #72, opened 2021-06-08
Keep PG password between release upgrades

skorobogatydmitry:
Change been checked this way:
```
$ helm upgrade --install --create-namespace -n sd-test yw .                                
Release "yw" does not exist. Installing it now.
NAME: yw
LAST DEPLOYED: Tue Jun  8 15:56:26 2021
NAMESPACE: sd-test
STATUS: deployed
REVISION: 1
TEST SUITE: None
$ kubectl -n sd-test get configmap/yw-yugaware-global-config -o 'jsonpath={.data.postgres_password}'
PR9sEIfC
$ helm upgrade --install --create-namespace -n sd-test yw .                                         
...
REVISION: 2
...
$ kubectl -n sd-test get configmap/yw-yugaware-global-config -o 'jsonpath={.data.postgres_password}'
PR9sEIfC
```
But `helm template` shows some random value, what should be ok.

### DIFF
--- a/stable/yugaware/templates/_helpers.tpl
+++ b/stable/yugaware/templates/_helpers.tpl
@@ -80,3 +80,19 @@ Validate Nginx SSL protocols
     {{- .Values.tls.sslProtocols -}}
   {{- end -}}
 {{- end -}}
+
+{{/*
+Get or generate PG password
+Source - https://github.com/helm/charts/issues/5167#issuecomment-843962731
+*/}}
+{{- define "getOrGeneratePassword" }}
+{{- $len := (default 8 .Length) | int -}}
+{{- $obj := (lookup "v1" .Kind .Namespace .Name).data -}}
+{{- if $obj }}
+{{- index $obj .Key -}}
+{{- else if (eq (lower .Kind) "secret") -}}
+{{- randAlphaNum $len | b64enc -}}
+{{- else -}}
+{{- randAlphaNum $len -}}
+{{- end -}}
+{{- end }}
--- a/stable/yugaware/templates/global-config.yaml
+++ b/stable/yugaware/templates/global-config.yaml
@@ -12,6 +12,6 @@ metadata:
     heritage: {{ .Values.helm2Legacy | ternary "Tiller" (.Release.Service | quote) }}
 data:
   postgres_user: "postgres"
-  postgres_password: "{{ b64enc (randAlphaNum 8) }}"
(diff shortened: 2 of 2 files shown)


iSignal (approved):
@skorobogatydmitry : nice, this solves a long standing annoyance. 

For the test above, can you clarify that you tested setting up a YW with charts code before this change and then upgrading the helm chart with this change included? It would be good to test that scenario. We expect no updates to happen in this case.

Also, does this trick work for helm template too? What does helm template show in the postgres_password config map entry on upgrade?

stable/yugaware/templates/_helpers.tpl
```diff
@@ -80,3 +80,19 @@ Validate Nginx SSL protocols
...
   {{- end -}}
 {{- end -}}
+
+{{/*
+Get or generate PG password
+Source - https://github.com/helm/charts/issues/5167#issuecomment-843962731
+*/}}
+{{- define "getOrGeneratePassword" }}
+{{- $len := (default 8 .Length) | int -}}
+{{- $obj := (lookup "v1" .Kind .Namespace .Name).data -}}
```
rtsisyk, line 90: Please lookup secret instead.

Here is an example:
```
{{- $secret_name := printf "%s-postgres" .Release.Name -}}
{{- $secret := lookup "v1" "Secret" .Release.Namespace $secret_name }}
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: {{ $secret_name }}
  labels:
    {{- include "chart.postgres.labels" . | nindent 4 }}
data:
  POSTGRES_DB: "{{ .Release.Name | b64enc }}"
  POSTGRES_USER: "{{ .Release.Name | b64enc }}"
  {{- if or (not $secret) (not $secret.data) }}
  POSTGRES_PASSWORD: {{ randAlphaNum 20 | b64enc }}
  {{- else }}
  POSTGRES_PASSWORD: {{ index $secret.data "POSTGRES_PASSWORD" }}
  {{ end }}

```

skorobogatydmitry:
> @skorobogatydmitry : nice, this solves a long standing annoyance.
> 
> For the test above, can you clarify that you tested setting up a YW with charts code before this change and then upgrading the helm chart with this change included? It would be good to test that scenario. We expect no updates to happen in this case.
> 
> Also, does this trick work for helm template too? What does helm template show in the postgres_password config map entry on upgrade?

Checked your scenario - it works as expected. But, old chart got upgraded, there will be wrong permanent password.

`helm template` shows random new value each time.

PR #73, opened 2021-06-08
Add PostgreSQL service

skorobogatydmitry:
Also moved statefulset to separated file.
Testing plan:
- install clear chart: `helm upgrade --install --create-namespace -n sd-test yw .`
  > No service at this point
- enable PG service: `helm upgrade --install --create-namespace -n sd-test --set service.postgres=true yw .`
  > Service yw-postgres appeared
- check psql connection:
  - add sidecar pod
  - exec to it
  - install psql: `yum install -y postgresql`
  - try `psql -h yw-postgres` - prompt for password
- upgrade release with postgres disabled: `helm upgrade --install --create-namespace -n sd-test --set yugaware.service.postgres=false yw .`
  > No postgres service at this point
- delete release: `helm -n sd-test delete yw`
  > No orphaned objects at this point


stable/yugaware/templates/service.yaml
```diff
@@ -40,199 +40,28 @@ spec:
...
   selector:
-    matchLabels:
-      app: {{ .Release.Name }}-yugaware
-  template:
-    metadata:
-{{- if .Values.yugaware.pod.annotations }}
-      annotations:
-{{ toYaml .Values.yugaware.pod.annotations | indent 8 }}
+    app: {{ .Release.Name }}-yugaware
+  type: ClusterIP
```
iSignal: let's make the service type and annotations configurable just like the other service

stable/yugaware/templates/service.yaml
```diff
@@ -40,199 +40,28 @@ spec:
...
-apiVersion: apps/v1
-kind: StatefulSet
+apiVersion: v1
+kind: Service
 metadata:
-  name: {{ .Release.Name }}-yugaware
+  name: {{ .Release.Name }}-postgres
+  {{- if .Values.yugaware.service.annotations }}
+  annotations:
+    {{ toYaml .Values.yugaware.service.annotations | indent 4 }}
```
iSignal: let's use postgres.service.annotations

stable/yugaware/values.yaml
```diff
@@ -44,6 +43,7 @@ yugaware:
     enabled: true
     ip: ""
     type: "LoadBalancer"
+    postgres: false # Expose internal postgres as an internal Service
```
iSignal: let's move this to a separate section like 

postgres;
  service:
    enabled: true

similar to the existing yugaware config

iSignal (commented):
Looks good overall. I can't see a diff between the new file and the old code - is it the same code in a different file? (it is better to check in formatting changes like this in a separate change so that it is easier to see the diff here).

stable/yugaware/templates/service.yaml
skorobogatydmitry: I wanted to keep this patch small as possible.
IMO adding same features as for YW makes it harder to maintain and remove in future.
But ok, let's make it a bit smarter.

skorobogatydmitry:
> Looks good overall. I can't see a diff between the new file and the old code - is it the same code in a different file? (it is better to check in formatting changes like this in a separate change so that it is easier to see the diff here).

Let me give an example why such changes worth to be in same patch along with the main feature.
E.g. now I want to don't make a loop and check chart manually. So I would add GH workflow with `helm template`. But the first comment about GH workflow in this patch will be "please do in a separate PR". And it makes it useless for the current work I need to complete.

skorobogatydmitry:
Last change checked by this: `helm upgrade --install --create-namespace -n sd-test --set postgres.service.enabled=true --set postgres.service.annotations.some=other --set postgres.service.annotations.someone=otherone --set postgres.service.type=LoadBalancer yw .`

stable/yugaware/values.yaml
```diff
@@ -56,6 +55,13 @@ yugaware:
       memory: 4Gi
   enableProxyMetricsAuth: true
 
+postgres: # Configure PostgreSQL part of the application
+  service:
+    enabled: false # Expose internal Postgres as a Service
+    annotations: {} # Additional Service annotations
+    type: ClusterIP # Service type
+    ip: "" # IP for LoadBalancer
```
```suggestion
## Configure PostgreSQL part of the application
postgres:
  service:
    ## Expose internal Postgres as a Service
    enabled: false
    ## Additional Service annotations
    annotations: {}
    ## Service type
    type: "ClusterIP"
    ## IP address for the LoadBalancer, works only if supported by the cloud provider
    ip: ""
```
Usually this is the convention followed by most of the charts.

bhavin192 (commented):
Mostly a formatting/convention change here. 
And when merging this PR the word "headless" can be removed from the commit message, because we are creating a ClusterIP service by default.

iSignal (approved):

PR #74, opened 2021-06-09
[charts] Added new k8s label topology.kubernetes.io/zone

### LINKED ISSUE #8586: [charts] Replace deprecated annotation failure-domain.beta.kubernetes.io/zone in yugabyte charts
The beta zone labels for zone/region have been deprecated in https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesiozone. 

We could perhaps specify both annotations until both are supported - for example, see https://github.com/Azure/AKS/issues/2188

@kkg-yb 

@jvigil-yugabyte : this probably requires your helm versioning changes to go in.

baba230896:
### Summary

- Earlier it was using this label `failure-domain.beta.kubernetes.io/zone` for the same purpose & it's deprecated.
- Now the YB charts have both labels in the node affinity, which means one should exist on the node to schedule a pod.
- `nodeSelectorTerms` is ORed, and `matchExpressions` is Anded expression.

### Test Plan:
Used this official YB doc to deploy the universe in a multi-zone scenario.
https://docs.yugabyte.com/latest/deploy/kubernetes/multi-zone/gke/helm-chart/

Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity

Fixes https://github.com/yugabyte/yugabyte-db/issues/8586

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -181,6 +181,11 @@ spec:
                 operator: In
                 values:
                 - {{ $root.Values.AZ }}
+            - matchExpressions:
+              - key: topology.kubernetes.io/zone
+                operator: In
+                values:
+                - {{ $root.Values.AZ }}
         {{ end }}
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:


iSignal:
What is the earliest k8s version where topology.kubernetes.io is understood?

baba230896:
@iSignal 
Kubernetes introduced the new labels (`topology...`) & deprecated the beta labels (`failure-domain.beta...`) in `v1.17`. 

Ref:
1. https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesiozone
2. https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.17.md#other-notable-changes-4
3. https://github.com/kubernetes/kubernetes/pull/81431

iSignal:
1. What would the impact be if someone is deploying this helm chart in version < 1.17? Is there any way to support both old and new labels?
2. Does the change above trigger a restart of the statefulset if no other changes are made to the chart?

baba230896:
> What would the impact be if someone is deploying this helm chart in version < 1.17? Is there any way to support both old and new labels?

The changes are backward compatible. The current implementation will schedule a pod if one of them (`topology...`/`failure-domain.beta.`) exists on the node. The `nodeSelectorTerms` does `ORed` operation, which means it schedule a pod on the node if that node has one of the labels defined in it. And if both of them won't exist, then the pod will be stuck in the pending state.

> Does the change above trigger a restart of the `statefulset` if no other changes are made to the chart?

Yes, it'll create a restart.

iSignal (approved):
Makes sense, thanks for the clarifications.

bhavin192 (approved):

PR #81, opened 2021-07-06
Add support for using external postgresql instances and supporting infrastructure.

kevincox:


rtsisyk (approved):
LGTM

stable/yugaware/templates/configs.yaml
```diff
@@ -34,8 +34,9 @@ data:
 
     db {
       default.host="{{ eq .Values.ip_version_support "v6_only" | ternary "[::1]" "127.0.0.1" }}"
-      default.port=5432
-      default.url="jdbc:postgresql://"${db.default.host}":"${db.default.port}"/"${POSTGRES_DB}
+      default.port={{ .Values.postgres.external.port }}
```
jvigil-yugabyte, line 38: Should we use this value only if the external host is supplied and otherwise use some default value? It may be a bit surprising that these values are used even if the external host is not supplied. And, same in the global-config.yaml where the external user, password, and dbname are used.

kevincox: I agree it is a little weird but it seemed like extra complexity. I added a warning in docs that you shouldn't touch `postgres.external` unless you set `postgres.host`. There are three main options.

1. As it is now, use the values as defaults but the users aren't allowed to (but not prevented from) changing them.
    - The simplest option.
2. Move the values directly into the `postgres` object and use it even for the local db.
    - It allows the user to change internal values which is unnecessary moving parts.
3. Only use these values conditionally.
    - This has a lot of code overhead.

jvigil-yugabyte: May be able to conditionally use the values without too much code overhead by making use of helm's ternary operator. The templating could be kept to one-line for each instance, at least

kevincox: I implemented it. I think it is unnecessarily complicated but it isn't too bad. Let me know what you think.

jvigil-yugabyte (approved):

PR #87, opened 2021-08-03
Add user-configurable volume annotations.

kevincox:
This allows configuring things such as automatic snapshots.

https://yugabyte.atlassian.net/browse/CLOUDGA-1243

### DIFF
--- a/stable/yugaware/values.yaml
+++ b/stable/yugaware/values.yaml
@@ -39,6 +39,7 @@ yugaware:
   replicas: 1
   storage: 100Gi
   storageClass: standard
+  storageAnnotations: {}
   multiTenant: false
   serviceAccount: yugaware
   serviceMonitor:
--- a/stable/yugaware/templates/volumes.yaml
+++ b/stable/yugaware/templates/volumes.yaml
@@ -3,6 +3,10 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: {{ .Release.Name }}-yugaware-storage
+{{- if .Values.yugaware.storageAnnotations }}
+  annotations:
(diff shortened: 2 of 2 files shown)


stable/yugaware/values.yaml
```diff
@@ -65,6 +65,8 @@ yugaware:
   # enables features specific for cloud deployments
   cloud:
     enabled: false
+  volume:
```
jvigil-yugabyte: Do we want to move `storage` and `storageClass` into this new volume object? It will cause some pain for upgrading, but will be better organization for the long term.

Also, can we rename `volume` -> `pvc`? This is more of a minor issue, but it's a bit confusing to use the term "volume" standalone because there is another k8s resource called `PersistentVolume`.

kevincox: I thought about that but decided to keep backwards compatibility. Maybe we can file a bug with a "breaking-change" label or something? Then next time we release a major version we can move them into here?

Alternatively I can name this `storageAnnotations` which is somewhat mismatched with `service` but and `pod` but matches the other storage options.

How about `persistentVolumeClaim` so that non-kube experts can still grok what this is about without looking up every acronym in the file. (Not that "volume" is a great name to start with...) I also wouldn't mind just calling it `storage` or `disk`.

jvigil-yugabyte: I'm happy with just `storageAnnotations` so it matches the other storage fields. We should also move it to line 42 so that it's colocated with the other storage values.

kevincox: Done!

jvigil-yugabyte (approved):
:+1:

PR #88, opened 2021-08-12
Fix user annotations.

kevincox:
The second annotations attribute was overriding the first and causing it to be ignored.


templates
```diff
@@ -0,0 +1 @@
+stable/yugaware/templates
```
jvigil-yugabyte: what is this? lgtm aside from this extra change

jvigil-yugabyte (approved):

PR #89, opened 2021-09-04
[PLAT-1576] Add CORS allowed origins list to parameters

skorobogatydmitry:
Without anything set - `helm template sd-test .`
```
....
                allowedOrigins = ["http://localhost"]
....
```

With the following overrides:
```
yugaware:
  additionAllowedCorsOrigins:
    - http://asd
    - dsa
```
and template `helm template sd-test . -f o.yaml` it comes to
```
....
allowedOrigins = ["http://asd","dsa","http://localhost"]
....
```

Format itself checked manually on cloud platform.

### DIFF
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -101,11 +101,7 @@ data:
       # CORS config
       cors {
         pathPrefixes = ["/"]
-{{- if .Values.tls.enabled }}
-        allowedOrigins = ["https://{{ .Values.tls.hostname }}"]
-{{- else }}
-        allowedOrigins = ["http://{{ .Values.tls.hostname }}"]
-{{- end }}
+        allowedOrigins = {{ include "allowedCorsOrigins" . }}
         # Server allows cookies/credentials to be sent with cross-origin requests
         supportsCredentials=true
         allowedHttpMethods = ["GET", "POST", "PUT", "OPTIONS", "DELETE"]
--- a/stable/yugaware/values.yaml
+++ b/stable/yugaware/values.yaml
@@ -64,8 +64,6 @@ yugaware:
   enableProxyMetricsAuth: true
   ## List of additional alowed CORS origins in case of complex rev-proxy
   additionAllowedCorsOrigins: []
-  #   - http://external.host.domain.local
-  #   - https://mylocalhost
   proxyEndpointTimeoutMs: 1 minute
   ## Enables features specific for cloud deployments
   cloud:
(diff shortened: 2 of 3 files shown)


stable/yugaware/values.yaml
```diff
@@ -62,8 +62,12 @@ yugaware:
       cpu: 2
       memory: 4Gi
   enableProxyMetricsAuth: true
+  ## List of additional alowed CORS origins in case of complex rev-proxy
+  additionAllowedCorsOrigins: []
```
danielisen, line 66: Can we add the teleport proxied platform addrs to this list? This was the issue I was running into yesterday

stable/yugaware/templates/configs.yaml
```diff
@@ -108,11 +108,7 @@ data:
       # CORS config
       cors {
         pathPrefixes = ["/"]
-{{- if .Values.tls.enabled }}
-        allowedOrigins = ["https://{{ .Values.tls.hostname }}"]
```
danielisen, line 112: Why removing the tls hostname? We need that + the teleport hostname used when proxying platform UIs

stable/yugaware/values.yaml
skorobogatydmitry: Yes, i;m working on this

jvigil-yugabyte: we shouldn't hardcode the value here, correct? instead, we'd want to add it as a values override when deploying the platform when teleport proxy addrs are required as an allowed origin

stable/yugaware/templates/configs.yaml
skorobogatydmitry: I don't remove it, just put to helpers to include there.
As Teleport Platform URL is dynamic (depends on cloud + network_id + region + platform_id), so I can't add it here.

stable/yugaware/values.yaml
skorobogatydmitry: Right.

skorobogatydmitry: This patch is in "Ready for review" state. So no changes expected, but someone asked for it.

skorobogatydmitry: So I'm working on this in Cloud infra.

stable/yugaware/values.yaml
```diff
@@ -62,8 +62,12 @@ yugaware:
...
+  #   - http://external.host.domain.local
+  #   - https://mylocalhost
```
jvigil-yugabyte: I think these comments (on lines 67-68) are a bit misleading. If the user uncomments these lines, the above empty list will get in the way of being able to use them. Would propose to remove these comments. Otherwise, lgtm.

jvigil-yugabyte (approved):

stable/yugaware/values.yaml
skorobogatydmitry: NP, will remove comments

PR #90, opened 2021-09-05
Add the option to provision certificates via cert-manager

ThoreKr:
Good evening,

we ran into issues with the helm generated certificates and weren't exactly happy with the way it handled certificate expiration.
Since we already had cert-manager in place and i thought there was an issue requesting support for it, here are the changes necessary to support it.

This pr allows to enable the use of cert-manager (version >=1.5) and will then issue certifificate ressources for the client, master and tserver.
By default it will bootstrap it's own selfsigned ca, but it is possible to override that and use another ca in the cluster. 

There are a few open questions which might need further discussion or feedback:
- Should the commits be squashed?
- Currently all certificates will all be issued by the same issuer. Would it make sense to allow to split the issuers for the nodes and the client certificates
- How does this change relate to https://github.com/yugabyte/yugabyte-db/issues/9443, more precisely, when the certificates will be renewed, will the nodes pick them up automatically or will it be necessary to somehow externally monitor these and lifecycle the statefulsets if necessary?
- How far should test results be reported?


stable/yugabyte/templates/certificates.yaml
```diff
@@ -0,0 +1,122 @@
...
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: yugabyte-ca
+  namespace: "{{ $root.Release.Namespace }}"
+spec:
+  isCA: true
+  privateKey:
+    algorithm: ECDSA
+    size: 521
```
jvigil-yugabyte: 512?

stable/yugabyte/templates/certificates.yaml
```diff
@@ -0,0 +1,122 @@
...
+    {{- end }}
+  secretName: yugabyte-{{ .name }}-tls
+  duration: {{ $root.Values.tls.certManager.certificateLifetime.duration | quote }}
+  renewBefore: {{ $root.Values.tls.certManager.certificateLifetime.renewBefore | quote }}
+  commonName: yugabyte-{{ .name }}
+  isCA: false
+  privateKey:
+    algorithm: ECDSA
+    encoding: PKCS8
+    size: 521
```
jvigil-yugabyte: 512?

stable/yugabyte/templates/certificates.yaml
```diff
@@ -0,0 +1,122 @@
...
+    {{- end }}
+  secretName:  {{ $root.Values.oldNamingStyle | ternary "yugabyte-tls-client-cert" (printf "%s-client-tls" (include "yugabyte.fullname" $root)) }}
+  duration: {{ $root.Values.tls.certManager.certificateLifetime.duration | quote }}
+  renewBefore: {{ $root.Values.tls.certManager.certificateLifetime.renewBefore | quote }}
+  commonName: yugabyte
+  isCA: false
+  privateKey:
+    algorithm: ECDSA
+    encoding: PKCS8
+    size: 521
```
jvigil-yugabyte: 512?

stable/yugabyte/templates/certificates.yaml
```diff
@@ -0,0 +1,122 @@
...
+      {{- include "yugabyte.applabel" ($appLabelArgs) | indent 6 }}
+      {{- include "yugabyte.labels" $root | indent 6 }}
+  issuerRef:
+    name:  {{ include "yugabyte.tls_cm_issuer" $root | quote }}
+    {{- if $root.Values.tls.certManager.useClusterIssuer }}
+    kind: ClusterIssuer
+    {{- else }}
+    kind: Issuer
+    {{- end }}
+  secretName: yugabyte-{{ .name }}-tls
```
jvigil-yugabyte: Can we make this support the new naming style (to allow multiple yugabyte universes to be deployed in the same namespace, if desired)?
```
{{ $root.Values.oldNamingStyle | ternary (printf "%s-yugabyte-tls-cert" .label) (printf "%s-%s-tls-cert" (include "yugabyte.fullname" $root) .label) }}
```

stable/yugabyte/templates/service.yaml
```diff
@@ -426,11 +426,37 @@ spec:
         {{- if $root.Values.tls.enabled }}
         - name: {{ $root.Values.oldNamingStyle | ternary (printf "%s-yugabyte-tls-cert" .label) (printf "%s-%s-tls-cert" (include "yugabyte.fullname" $root) .label) }}
           secret:
+            {{- if not $root.Values.tls.certManager.enabled }}
             secretName: {{ $root.Values.oldNamingStyle | ternary (printf "%s-yugabyte-tls-cert" .label) (printf "%s-%s-tls-cert" (include "yugabyte.fullname" $root) .label) }}
+            {{- else }}
+            secretName: yugabyte-{{ .name }}-tls
```
jvigil-yugabyte: After modifying the cert secretName to match the existing secretName (to support new naming convention), we'll have to rework this logic slightly.

jvigil-yugabyte (changes requested):
Thank you for adding this functionality!

stable/yugabyte/templates/certificates.yaml
ThoreKr: For some curious reason for elliptic curves it has to be 521. 
On that note, would it make sense to expose key size an algorithm to the values.yaml?

stable/yugabyte/templates/certificates.yaml
ThoreKr: Would it still make sense to generate separate certificates for masters and tservers?

Initially I tried to find a way to prevent a way that every node also has access to every other nodes certificate and eventually settled to split masters and tservers with one certificate for each one and the replicas as alternative names.

stable/yugabyte/templates/certificates.yaml
jvigil-yugabyte: Huh, did not realize this. Perhaps it would make sense to expose these options.

stable/yugabyte/templates/service.yaml
```diff
@@ -426,11 +426,37 @@ spec:
         {{- if $root.Values.tls.enabled }}
         - name: {{ $root.Values.oldNamingStyle | ternary (printf "%s-yugabyte-tls-cert" .label) (printf "%s-%s-tls-cert" (include "yugabyte.fullname" $root) .label) }}
           secret:
+            {{- if not $root.Values.tls.certManager.enabled }}
             secretName: {{ $root.Values.oldNamingStyle | ternary (printf "%s-yugabyte-tls-cert" .label) (printf "%s-%s-tls-cert" (include "yugabyte.fullname" $root) .label) }}
+            {{- else }}
+            secretName:  {{ $root.Values.oldNamingStyle | ternary (printf "yugabyte-%s-tls" $service.label) (printf "%s-%s-tls" (include "yugabyte.fullname" $root) $service.label) }}
```
jvigil-yugabyte: What do you think about re-using the same `secretName` regardless of whether or not `certManager.enabled` is true?

stable/yugabyte/templates/certificates.yaml
jvigil-yugabyte: For now, we can match the current behavior of generating separate certificates for masters and tservers. This will make it easier to change in the future (all at once) if necessary.

jvigil-yugabyte (approved):
LGTM :)

stable/yugabyte/templates/certificates.yaml
```diff
@@ -0,0 +1,123 @@
...
+  usages:
+    - server auth
+    - client auth
+  # At least one of a DNS Name, URI, or IP address is required.
+  dnsNames:
+  {{- $replicas := (eq .name "yb-masters") | ternary $root.Values.replicas.master $root.Values.replicas.tserver -}}
+  {{- range $index := until ( int ( $replicas ) ) }}
+  {{- $nodeOldStyle := printf "%s-%d.%s.%s.svc.%s" $service.label $index $service.name $root.Release.Namespace $root.Values.domainName }}
+  {{- $nodeNewStyle := printf "%s-%s-%d.%s-%s.%s.svc.%s" (include "yugabyte.fullname" $root) $service.label $index (include "yugabyte.fullname" $root) $service.name $root.Release.Namespace $root.Values.domainName }}
+  {{- $node := $root.Values.oldNamingStyle | ternary $nodeOldStyle $nodeNewStyle }}
```
iSignal, line 83: So it seems like we are generating a single server cert that goes on all tservers and its SAN has all the DNS names of all tserver (and another with all masters). This is different behavior from the older method but should be ok.

I am not clear how this gets mounted into the pod as `node.DNS.crt` and `node.DNS.key` like the old cert used to?

stable/yugabyte/templates/certificates.yaml
```diff
@@ -0,0 +1,123 @@
...
+  issuerRef:
+    name:  {{ include "yugabyte.tls_cm_issuer" $root | quote }}
+    {{- if $root.Values.tls.certManager.useClusterIssuer }}
+    kind: ClusterIssuer
+    {{- else }}
+    kind: Issuer
+    {{- end }}
+  secretName: {{ $root.Values.oldNamingStyle | ternary (printf "%s-yugabyte-tls-cert" $service.label) (printf "%s-%s-tls-cert" (include "yugabyte.fullname" $root) $service.label) }}
+  duration: {{ $root.Values.tls.certManager.certificates.duration | quote }}
+  renewBefore: {{ $root.Values.tls.certManager.certificates.renewBefore | quote }}
```
iSignal, line 67: It seems like cert manager will update this cert at renewal time, which presumbly gets reflected on files within the pod. Right now, there is no automatic reload of the certs, so certs are not actually reloaded until a restart happens. Even when https://github.com/yugabyte/yugabyte-db/issues/9443 is implemented, an explicit signal or RPC will be required to achieve this reload. To do so automatically would probably require some kind of a 'hook'

I would add this information as a comment in values.yaml to make it clear that renewal period can be set but has to followed up with an explicit restart.

stable/yugabyte/templates/certificates.yaml
```diff
@@ -0,0 +1,123 @@
...
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ $root.Values.oldNamingStyle | ternary "yugabyte-ca" (printf "%s-ca" (include "yugabyte.fullname" $root)) }}
+  namespace: "{{ $root.Release.Namespace }}"
+spec:
+  isCA: true
+  privateKey:
```
iSignal, line 21: This may be a naive question but I assume that once created as part of the initial `helm install`, this Certificate is not updated by cert-manager on `helm upgrade`s because none of its parameters have changed?

iSignal (commented):
Thanks for this contribution to yugabyte-db. This is a very polished PR and thanks for maintaining compat with existing params and the old method.

Overall, it looks good. I have a few minor clarifying comments.

To answer your questions,
* Squashed commits is preferred, yes
* Splitting node vs client certs could make sense in some cases. For example, when clients alone are exposed out through a LB publicly, one may choose to use "properly signed" certs for the certs that are placed by the `--certs_for_client_dir` parameter and the corresponding client auth cert. The issuer for `--certs_for_client_dir` has to match the issuer for the client auth cert (which is really a postgres client cert). Right now, we don't split out the node certs for `--certs_dir` and client facing certs for `--certs_for_client_dir`, so supporting different issuers here will require more work.
* Cert file changes are not reloaded automatically, so renewal will require an externally coordinated restart. The linked issue will eventually make it possible to achieve this via a SIGHUP or RPC call but will still not be automatic (postgres follows a similar policy of using SIGHUP to reload certs, so we could investigate how postgres-based helm chartrs handle cert-manager renewal.
* For testing, I would use `helm template` to verify that things are unchanged when cert manager is disabled. When it is enabled, I would test `helm install` followed by a `helm upgrade` that increases tserver replica count. Testing client certs is a bit more effort, let us know if you are interested but it is not necessary.

stable/yugabyte/templates/certificates.yaml
ThoreKr: Indeed it generates a certificate for the tservers and one for the masters. Initially I had hoped to be able to fully get around having shared certificates but couldn't figure out how to get that aligned with statefulset replicas. The closest I could get was to split between masters and tservers.

The mapping to the `node.DNS.crt/key` mounts happens in the volume definition https://github.com/yugabyte/charts/pull/90/files#diff-1c1cc3ecca03b17a073b1ab59f5e61d1c8c0954950627d0613e46ca157a1db5fR431 as a loop over the items. Surprisingly it is possible to mount the same key multiple times (cert-manager has a fixed schema for naming the keys in the secret).

stable/yugabyte/templates/certificates.yaml
ThoreKr: This is indeed an issue, which we could reproduce with certificates which had a 2h lifetime.

I will add a comment to the `values.yaml`

stable/yugabyte/templates/certificates.yaml
ThoreKr: cert-manager has a couple of restrictions on how it will update the secrets containing the certificates, whenever the Certificate resource is updated. But as long as it doesn't detect any changes or reasons to re-issue a certificate the actual secrets will remain untouched.

stable/yugabyte/templates/certificates.yaml
iSignal: Thanks for the pointer, it makes sense.

iSignal:
Thanks for the responses. Please update the test plan to reflect the test cases above and add the comment in values.yaml mentioned above and we should be good to merge it in.

ThoreKr:
Alright. Thanks fr the feedback.

Commits are squashed.

Test scenarios performed:
  - helm template on this branch with defaults `values.yaml` and master - diff found no changes
  - deployment with cert-manager enabled, no further modifications
    - scaled up tservers from 3 to 4. certificate got reissued with additional dnsname und picked up by the new tserver
    - Set certificate lifetime to 2h --> cluster will break after that time
   - deployment with cert-manager enabled, and separate cluster issuer

PR #91, opened 2021-09-18
Added option to set custom annotations to the StatefulSet

jannh:
Added the option to set custom annotations to the StatefulSet.
Intended first use on our cluster would be to add an annotation for wave ( https://github.com/wave-k8s/wave ) to be able to trigger a rolling-update of the pods, when the SSL certificates of the cluster are renewed. I'm sure, there are lots of further use cases.

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -112,6 +112,10 @@ metadata:
   labels:
     {{- include "yugabyte.applabel" ($appLabelArgs) | indent 4 }}
     {{- include "yugabyte.labels" $root | indent 4 }}
+  {{ if $root.Values.statefulSetAnnotations }}
+  annotations:
+{{ toYaml $root.Values.statefulSetAnnotations | indent 4}}
+  {{ end }}
 spec:
   serviceName: {{ $root.Values.oldNamingStyle | ternary .name (printf "%s-%s" (include "yugabyte.fullname" $root) .name) | quote }}
   podManagementPolicy: {{ $root.Values.PodManagementPolicy }}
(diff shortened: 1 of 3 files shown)


jannh:
@iSignal @jvigil-yugabyte any suggestions for improvements?

stable/yugabyte/templates/service.yaml
```diff
@@ -112,6 +112,10 @@ metadata:
   labels:
     {{- include "yugabyte.applabel" ($appLabelArgs) | indent 4 }}
     {{- include "yugabyte.labels" $root | indent 4 }}
+  {{ if $root.Values.statefulSetAnnotations }}
```
jvigil-yugabyte: nit: we should trim the whitespace here:
```
  {{- if $root.Values.statefulSetAnnotations }}
```

stable/yugabyte/templates/service.yaml
```diff
@@ -112,6 +112,10 @@ metadata:
...
+  annotations:
+{{ toYaml $root.Values.statefulSetAnnotations | indent 4}}
```
jvigil-yugabyte: nit: space after the 4

stable/yugabyte/templates/service.yaml
```diff
@@ -112,6 +112,10 @@ metadata:
...
+  {{ end }}
```
jvigil-yugabyte: nit: we should trim whitespace here also

jvigil-yugabyte (commented):
Also, there is a `.Values.commonLabels` helm chart value, though it doesn't appear in our values.yaml file.

jvigil-yugabyte (approved):
LGTM, thank you!

PR #93, opened 2021-09-23
Make PodDisruptionBudget configurable.

kevincox:
The old default of `maxUnavailable: 0` doesn't make sense because a StatefulSet doesn't allow extra replicas to be scheduled (not that kube tries to upsize collections anyways). This effectively means that the pods will never be allowed to be evicted which just results in inefficient use of a kube cluster and block updates. This makes the PDB configurable to allow testing of a better configuration.

Likely this will be updated to a different default value once it has been tested.

### DIFF
--- a/stable/yugaware/templates/pdb.yaml
+++ b/stable/yugaware/templates/pdb.yaml
@@ -3,7 +3,7 @@ kind: PodDisruptionBudget
 metadata:
   name: {{ .Release.Name }}-yugaware-pdb
 spec:
-  maxUnavailable: 0
+  maxUnavailable: {{ .Values.yugaware.podDisruptionBudget.maxUnavailable | toJson }}
   selector:
     matchLabels:
       app: {{ .Release.Name }}-yugaware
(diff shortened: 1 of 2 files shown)


stable/yugaware/templates/pdb.yaml
```diff
@@ -3,7 +3,7 @@ kind: PodDisruptionBudget
 metadata:
   name: {{ .Release.Name }}-yugaware-pdb
 spec:
-  maxUnavailable: 0
+  maxUnavailable: {{ .Values.yugaware.podDisruptionBudget.maxUnavailable | toJson }}
```
Arnav15, line 6: Do you need the toJson? Isn't an int value just cleanly applyable to the template?

Arnav15 (approved):

stable/yugaware/templates/pdb.yaml
kevincox: It isn't strictly required but makes errors in the values provided much cleaner errors as the value is restricted to this attribute and doesn't spew out random text that is parsed as yaml.

PR #96, opened 2021-10-21
[yugabyte] Add liveness check based on disk IO

### LINKED ISSUE #10345: [Platform][K8S] Yugabyte DB Pod should crash should any of the disks become inaccessible
We need a Readiness Probe and Liveness probe to handle disk failure scenarios:
- Readiness Probe that fails if t-server fails to come up when bringing up a DB pod
- Liveness Probe that results in force restart of DB Pod If any of the disks become inaccessible while DB pod is running

bhavin192:
- In some Kubernetes setups it is possible that underlying disk dies and writes from TServer fail but this failure is not detected immediately, instead it is detected only during next write operation by TServer.
- This adds a simple liveness check to the database pods to check if the disk is readable and writable.

If the probe fails 3 times in a row, it will be marked as failed. The probe is checked every 10 seconds, so this will take 30 seconds in total.

### Scenarios tested:

- Installed the chart using latest master.
- Upgraded it with these changes.
- YB-Master and YB-TServer pods run as expected.
- Tried the storage.ephemeral and skipHealthChecks flags as well, the
  liveness probe is not added in case of these flags.

Fixes https://github.com/yugabyte/yugabyte-db/issues/10345

### DIFF
--- a/stable/yugabyte/templates/_helpers.tpl
+++ b/stable/yugabyte/templates/_helpers.tpl
@@ -106,6 +106,16 @@ Get YugaByte fs data directories.
 {{- end -}}
 
 {{/*
+Get files from fs data directories for readiness / liveness probes.
+*/}}
+{{- define "yugabyte.fs_data_dirs_probe_files" -}}
+  {{- range $index := until (int (.count)) -}}
+    {{- if ne $index 0 }} {{ end }}"/mnt/disk{{ $index -}}/disk.check"
+  {{- end -}}
+{{- end -}}
+
+
+{{/*
 Generate server FQDN.
 */}}
 {{- define "yugabyte.server_fqdn" -}}
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -229,6 +229,18 @@ spec:
                       cp /home/yugabyte/bin/log_cleanup.sh /mnt/disk0/yb-data/scripts;
                     fi;
                   fi
+        {{- if (and (not $root.Values.storage.ephemeral) (not $service.skipHealthChecks)) }}
+        livenessProbe:
+          exec:
+            command:
+            - bash
+            - -c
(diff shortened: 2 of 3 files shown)


stable/yugabyte/templates/service.yaml
```diff
@@ -229,6 +229,18 @@ spec:
                       cp /home/yugabyte/bin/log_cleanup.sh /mnt/disk0/yb-data/scripts;
```
iSignal, line 229: are you thinking that this serves as a sufficient preflight check? it is reasonable but if so, we should comment explicitly because it is hard to understand for future devs. we could also explicitly `touch` the file anyway just in case this log cleanup code changes later.

stable/yugabyte/templates/_helpers.tpl
```diff
@@ -105,6 +105,16 @@ Get YugaByte fs data directories.
   {{- end -}}
 {{- end -}}
 
+{{/*
+Get files from fs data directories for readiness / liveness probes.
+*/}}
+{{- define "yugabyte.fs_data_dirs_probe_files" -}}
+  {{- range $index := until (int (.count)) -}}
+    {{- if ne $index 0 }} {{ end }}/mnt/disk{{ $index -}}/disk.check
```
iSignal: if we are going to pass these to touch, we should quote them in case they have spaces in them

iSignal (changes requested):

stable/yugabyte/templates/service.yaml
bhavin192: No, I don't think the postStart is correct place for this check, though it serves as yet another place which is doing disk related operations, and fail in case of issues. 

The container command and this postStart command start asynchronously, so it is possible that our database process starts first and then fails.

Like Jason was saying, we should have a preflight check for this.

bhavin192: I have added a new preflight check for this.

iSignal (approved):

PR #100, opened 2021-11-19
Use policy/v1 of PodDisruptionBudget when it is available

### LINKED ISSUE #10597: [Platform][k8s] Migrate PodDisruptionBudget to use newer API version
The API version is pointing to policy/v1beta1 which won't be available in v1.25 and deprecated. Update clients to use policy/v1 API version of PodDisruptionBudget.

Note - The policy/v1beta1 API version of PodDisruptionBudget will no longer be served in v1.25. Migrate manifests and API clients to use the policy/v1 API version, available since v1.21.

```
$ k get poddisruptionbudgets yb-tserver-pdb -o yaml -n yb-demo

apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
annotations:
meta.helm.sh/release-name: yb-demo
meta.helm.sh/release-namespace: yb-demo
creationTimestamp: "2021-10-20T22:35:45Z"
generation: 1
labels:
app.kubernetes.io/managed-by: Helm
name: yb-tserver-pdb
namespace: yb-demo
resourceVersion: "11057"
selfLink: /apis/policy/v1/namespaces/yb-demo/poddisruptionbudgets/yb-tserver-pdb
uid: 173aa017-e848-4ab3-9baf-9098ed9b8faf
spec:
maxUnavailable: 1    
selector:
matchLabels:
app: yb-tserver
status:
```

API Deprecation guide -  https://kubernetes.io/docs/reference/using-api/deprecation-guide/

bhavin192:
The policy/v1beta1 of PodDisruptionBudget has been deprecated and will be removed in Kubernetes version 1.25. With this change we use policy/v1 of PodDisruptionBudget if it is available in the cluster or else we fallback to policy/v1beta1 for both the charts yugabyte and yugaware.

In case of `helm template` command, it will always report that the policy/v1/PodDisruptionBudget is missing unless we provide the `--api-versions policy/v1/PodDisruptionBudget` argument to it. The only place where `helm template` is used, is the GKE marketplace's deployer image, where don't have control over the exact `helm template` command. In that case the pdbPolicyVersionOverride value of yugaware chart can be used to explicitly set the APIVersion.

### Scenarios tested:

1. Installed both charts on Kubernetes 1.20, and checked the APIVersion of PDB resources
   ```
   $ kubectl get pdb -n bg-dep \
       -ojsonpath='{range .items[*]}{.metadata.name}{" "}{.apiVersion}{"\n"}{end}'
     yb-master-pdb policy/v1beta1
     …
   ```

2. Upgraded the Kubernetes cluster to 1.21, and again checked the APIVersion of PDB resources with above command.
   ```
   yb-master-pdb policy/v1
   ```
3. Ran `helm upgrade` for both the chart, and checked the APIVersion again. The upgrade command was successful, and APIVersion remained policy/v1.

4. Tried the following template commands, which yield expected results.
   ```
   helm template ./yugabyte => uses v1beta1
   helm template ./yugaware => uses v1beta1
   helm template ./yugabyte --api-versions policy/v1/PodDisruptionBudget => uses v1
   helm template ./yugaware --api-versions policy/v1/PodDisruptionBudget => uses v1
   helm template ./yugaware --set pdbPolicyVersionOverride=v1 => uses v1
   ```

Fixes https://github.com/yugabyte/yugabyte-db/issues/10597

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -478,7 +478,15 @@ spec:
         {{- end }}
 {{- if eq $root.Values.isMultiAz false }}
 ---
+{{/*
+TODO: switch to policy/v1 completely when we stop supporting
+Kubernetes versions < 1.21
+*/}}
+{{- if $root.Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
+{{- end }}
 kind: PodDisruptionBudget
 metadata:
   name: {{ $root.Values.oldNamingStyle | ternary (printf "%s-pdb" .label) (printf "%s-%s-pdb" (include "yugabyte.fullname" $root) .name) }}
--- a/stable/yugaware/templates/pdb.yaml
+++ b/stable/yugaware/templates/pdb.yaml
@@ -1,4 +1,14 @@
+{{/*
+TODO: switch to policy/v1 completely when we stop supporting
+Kubernetes versions < 1.21
+*/}}
+{{- if .Values.pdbPolicyVersionOverride }}
+apiVersion: policy/{{ .Values.pdbPolicyVersionOverride }}
+{{- else if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
+{{- end }}
 kind: PodDisruptionBudget
 metadata:
   name: {{ .Release.Name }}-yugaware-pdb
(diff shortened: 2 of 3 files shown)


jvigil-yugabyte (approved):

jvigil-yugabyte:
One thing to check -- is it possible to upgrade from a v1beta1 PDB to a v1 PDB?

bhavin192:
@jvigil-yugabyte yes, it is. When the Kubernetes cluster administrator updates the cluster (like we did in the step number 2), Kubernetes will take care of making required changes to update the resource (I'm not going too deep about this explanation right now). So, the users have to make sure that they update their chart installation before they update to Kubernetes version 1.25.

PR #101, opened 2021-11-23
[Platform] Service Account related changes for YB platform GKE app

baba230896:
### Summary
YB platform charts are using the `{{ .Release.Name }}` instead of `yugaware.serviceAccount` of `values.yaml` for SA mentioned in STS.
The YB Platform GKE app creates the service account and attaches the RBAC permissions provided in `schema.yaml`.
As per the YB Platform GKE app requirements, we have modified the charts by which we can skip the SA creation by providing the SA name externally.
These changes won't impact the existing installation because the charts never used the `yugaware.serviceAccount`.

- Service account can be pass-through using the `--set`. It'll help in the YB Platform GKE app. 

### Test

- Compared the output of `helm template` with and without service account.
```bash
helm template yw-test stable/yugaware/ --set yugaware.serviceAccount=test-sa
helm template yw-test stable/yugaware/ 
```

### DIFF
--- a/stable/yugaware/templates/rbac.yaml
+++ b/stable/yugaware/templates/rbac.yaml
@@ -1,7 +1,8 @@
+{{ if not .Values.yugaware.serviceAccount }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ .Release.Name }}
+  name: {{ .Values.yugaware.serviceAccount | default .Release.Name }}
   labels:
     k8s-app: yugaware
     kubernetes.io/cluster-service: "true"
@@ -10,6 +11,7 @@ metadata:
   annotations:
 {{ toYaml .Values.yugaware.serviceAccountAnnotations | indent 4 }}
 {{- end }}
+{{ end }}
 {{- if .Values.rbac.create }}
 {{- if .Values.ocpCompatibility.enabled }}
 ---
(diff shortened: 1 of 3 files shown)


stable/yugaware/templates/rbac.yaml
```diff
@@ -1,7 +1,8 @@
+{{ if not .Values.yugaware.serviceAccount }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ .Release.Name }}
+  name: {{ .Values.yugaware.serviceAccount | default .Release.Name }}
```
bhavin192: This won't be needed here I think. As this ServiceAccount is created only when the yugaware.serviceAccount is not set.

stable/yugaware/values.yaml
```diff
@@ -41,7 +41,10 @@ yugaware:
   storageClass: ""
   storageAnnotations: {}
   multiTenant: false
-  serviceAccount: yugaware
+  # Providing service account externally won't create new SA.
+  # It just attaches the clusterrole to it. 
+  # Helpful in Yugabyte Platform GKE App.
```
```suggestion
  ## Name of existing ServiceAccount. When provided, the chart won't create a ServiceAccount.
  ## It will attach the required RBAC roles to it. 
  ## Helpful in Yugabyte Platform GKE App. 
```

The usual convention is to use two hashes before documentation comments in the values.yaml file.

bhavin192 (commented):
Can you update the commit message with details about the change in value of `yugaware.serviceAccount`. Mention that it was not used before, and hence this change won't affect any existing installation.

Also, mention how it is used in GKE app. i.e. GKE application creates the ServiceAccount for us, and passes it on to the chart.

stable/yugaware/templates/rbac.yaml
baba230896: Make sense.

bhavin192 (approved):

PR #102, opened 2021-12-08
[yugabyte] Added static IP for load balancer service

### LINKED ISSUE #10804: [Platform][k8s] Allow Specifying loadBalancerIP during helm install to get static IP
The ask is to add loadBalancerIP to yugabyte helm charts values.yaml as yugaware helm chart already allows [here](https://github.com/yugabyte/charts/blob/9248b65d52c64f5e45b4d9e3afa8f38c9734c930/stable/yugaware/templates/service.yaml#L41).

Additional context:
This is for the external load balancers. When using the load balancer service type, an external managed load balancer is created in the cloud hosting k8 cluster. Enabling helm chart config with loadBalancerIP field allows installing helm chart with providing static IP for LoadBalancer, otherwise, LoadBalancer is set up with an ephemeral IP address. 

GKE k8 yb-master-ui and yb-tserver-service LoadBalancer services created with Internal TCP/UDP load balancer using ephemeral IP due to annotation load-balancer-type: "Internal".  Clients can connect to this load balancer only internally within the cluster or region.

baba230896:
### Summary

- Added static IP support for load balancer service.
- Using this, we can associate YB Master UI and YB Tserver with reserved static IP.

### Testing

- I have passed the `values.yaml`, which has static IP, to `helm install` command with version - `2.9.1.0-b140`.
```bash
helm install yb-test stable/yugabyte/ -n test -f stable/yugabyte/values.yaml 
```

fix: https://github.com/yugabyte/yugabyte-db/issues/10804

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -100,6 +100,7 @@ spec:
     {{- include "yugabyte.appselector" ($appLabelArgs) | indent 4 }}
   type: {{ $endpoint.type }}
   externalTrafficPolicy: {{ $endpoint.externalTrafficPolicy | default "Cluster" }}
+  loadBalancerIP: {{ $endpoint.loadBalancerIP }}
 {{- end}}
 {{- end}}
 {{ end }}
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -100,7 +100,9 @@ spec:
     {{- include "yugabyte.appselector" ($appLabelArgs) | indent 4 }}
   type: {{ $endpoint.type }}
   externalTrafficPolicy: {{ $endpoint.externalTrafficPolicy | default "Cluster" }}
+  {{- if $endpoint.loadBalancerIP }}
   loadBalancerIP: {{ $endpoint.loadBalancerIP }}
+  {{- end }}
 {{- end}}
 {{- end}}
 {{ end }}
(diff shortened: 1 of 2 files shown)


bhavin192 (approved):

stable/yugabyte/templates/service.yaml
```diff
@@ -99,6 +99,7 @@ spec:
   selector:
     {{- include "yugabyte.appselector" ($appLabelArgs) | indent 4 }}
   type: {{ $endpoint.type }}
+  loadBalancerIP: {{ $endpoint.loadBalancerIP }}
```
jvigil-yugabyte, line 104: Does this affect upgrades? For example, if it is not set in a previous version, then we upgrade, will the upgrade succeed?

To protect against accidental chart changes, we could modify this to only specify `loadBalancerIP` if the `$endpoint.loadBalancerIP` value is set.

baba230896: > To protect against accidental chart changes, we could modify this to only specify loadBalancerIP if the $endpoint.loadBalancerIP value is set.

Yeah, Make sense.

baba230896:
@jvigil-yugabyte Could you please take a look at it whenever you've time?

jvigil-yugabyte (approved):
:+1:

stable/yugabyte/values.yaml
```diff
@@ -92,6 +92,7 @@ serviceEndpoints:
     ## Sets the Service's externalTrafficPolicy
     # externalTrafficPolicy: ""
     app: "yb-master"
+    loadBalancerIP: ""
```
```suggestion
    # loadBalancerIP: ""
```
How about keeping this commented out?

stable/yugabyte/values.yaml
```diff
@@ -100,6 +101,7 @@ serviceEndpoints:
     ## Sets the Service's externalTrafficPolicy
     # externalTrafficPolicy: ""
     app: "yb-tserver"
+    loadBalancerIP: ""
```
```suggestion
    # loadBalancerIP: ""
```

PR #103, opened 2021-12-14
[yugabyte] Make pod annotations, env, affinity configurable

### LINKED ISSUE #10603: [Platform][k8s] Support publishing NODE_IP instead of POD_IP as the broadcast address
Jira Link: [DB-1214](https://yugabyte.atlassian.net/browse/DB-1214)
### Description

We have a ticket opened by a customer, regarding their k8's helm chart. 2243
My understanding is in our Helm chart we use Pod IP while deploying the cluster. They want Node IP to be used instead, since that is what will be exposed to the applications.

So the change has to be done in Helm chart from our end, so that when ever they deploy it, NodeIP is used.
They have attached a zip of their helm chart files, there are a lot of changes, but we are only interested in the changes related to NODE_IP in file values.yaml. Updated the ticket with this info, as well as adding a link to a slack conversation that took place recently.

yb.zip is available on the lincoln server under /cases/2243.

bhavin192:
This also makes it possible to change externalTrafficPolicy for the
ServiceEndpoints, fallbacks to `Cluster` (Kubernetes default). Adds
ability to override the server_broadcast_address for TServer pods.

Adding two new sections (maps) to values.yaml as master and
tserver. Because maps are easy to override than lists (i.e. the
Services list we have). In the case of lists, one has to write the
whole list in their values file to override one key out of it. Maps
are merged recursively and hence it is possible to override a single
key.

1. The `master.affinity` and `tserver.affinity` can be used to
   override existing podAntiAffinity configuration as well as to add
   other affinity configuration like nodeAffinity, podAffinity.

2. `master.extraEnv` and `tserver.extraEnv` add given environment list
   to the respective pods.

3. `master.podAnnotations` and `tserver`podAnnotations` add given
   annotations to the pods. Useful in the setups where other tools depend
   on annotations to discover pods, for example, Prometheus can use these
   to monitor the pods.

4. `tserver.serverBroadcastAddress` can be used to override the default
   value of the `--server_broadcast_addresses` flag of TServer. Preflight
   checks are not done on the given `serverBroadcastAddress` as the value
   is mostly likely calculated when the pod runs.

   When tls.enabled is set along with this value, the
   `--cert_node_filename` flag with correct node name is added. This
   ensures that the yb-tserver process is able to find the correct
   certificate files generated by the chart.

Scenarios tested:

1. Ran helm template with the above change and without the change,
   only service's externalTrafficPolicy changes. Applying this doesn't
   cause any pod restart (tried with helm upgrade).
2. Installed the chart from master, and then set the values of
   podAnnotations, affinity, extraEnv, serverBroadcastAddress. The
   pods have all the values correctly set after the upgrade.
3. Used a chart having these changes with 2.11.0.0-b7 of
   platform. Added overrides at the AZ level when creating a new
   provider. It was able to spin up a universe successfully with this
   chart.

Fixes https://github.com/yugabyte/yugabyte-db/issues/10603

For the reviewer:
- We are skipping the preflight checks for the serverBroadcastAddress passed by the user, as it can be a environment variable or it can be something to which we might not be able to bind as well. 
- This change is using `--cert_node_filename` flag, which is not documented yet, it was introduced in https://github.com/yugabyte/yugabyte-db/commit/08d5ccbbc9213198099b926764ed2db121fe6af4

### DIFF
--- a/stable/yugabyte/values.yaml
+++ b/stable/yugabyte/values.yaml
@@ -89,12 +89,16 @@ domainName: "cluster.local"
 serviceEndpoints:
   - name: "yb-master-ui"
     type: LoadBalancer
+    ## Sets the Service's externalTrafficPolicy
+    # externalTrafficPolicy: ""
     app: "yb-master"
     ports:
       http-ui: "7000"
 
   - name: "yb-tserver-service"
     type: LoadBalancer
+    ## Sets the Service's externalTrafficPolicy
+    # externalTrafficPolicy: ""
     app: "yb-tserver"
     ports:
       tcp-yql-port: "9042"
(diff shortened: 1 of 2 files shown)


stable/yugabyte/values.yaml
```diff
@@ -217,6 +221,69 @@ affinity: {}
...
+  ## Ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#envvar-v1-core
+  ## Example:
+  # extraEnv:
+  # - name: NODE_IP
+  #   valueFrom:
+  #     fieldRef:
+  #       fieldPath: status.hostIP
+  extraEnv: []
+
+  ## Annotations to be added to the Master pods.
```
jvigil-yugabyte: Master -> TServer

jvigil-yugabyte (approved):

PR #113, opened 2022-02-23
[CLOUDGA-4518] Add option to proxy DB requests through nginx directly

danielisen:


stable/yugaware/templates/configs.yaml
```diff
@@ -176,6 +176,18 @@ data:
       location / {
         proxy_pass http://{{ eq .Values.ip_version_support "v6_only" | ternary "[::1]" "127.0.0.1" }}:9000;
       }
+{{- if .Values.nginx.db_node_proxy_enabled }}
+      location ~ "^/universes/.+/proxy/(.+):(7000|9000|9300|12000|13000)/(metrics|prometheus-metrics)$" {
+        proxy_pass "http://$1:$2/$3$is_args$args";
+        sub_filter "http://" "/proxy/";
```
kevincox: Can you explain why we need to do this filtering? I don't think the metrics endpoints should be returning any URLs.

danielisen: was implementing a more generic solution that also worked for going to the master/tserver UI before, but now that we've restricted the path to only metrics endpoints it is a moot point

stable/yugaware/templates/configs.yaml
```diff
@@ -176,6 +176,12 @@ data:
       location / {
         proxy_pass http://{{ eq .Values.ip_version_support "v6_only" | ternary "[::1]" "127.0.0.1" }}:9000;
       }
+{{- if .Values.nginx.db_node_proxy_enabled }}
+      location ~ "^/universes/.+/proxy/(.+):(7000|9000|9300|12000|13000)/(metrics|prometheus-metrics)$" {
```
kevincox: I think we should validate that this is an IP and block IPs in the 169.254 range as these are used by cloud metadata server and can contain auth credentials.

kevincox (approved):

PR #117, opened 2022-02-28
[Backport 2.6][yugabyte] Create a symlink to cores directory based on core_pattern

bhavin192:
Original commit: 64c4c8f9b3a28a33d3bf7d1b4672e311b081baad

Modifies the postStart hook to create a symlink to /mnt/disk0/cores at
the directory being used by kernel.core_pattern. This is useful when
the core_pattern's value points to some absolute path like
/var/lib/cores instead of a relative path and the path doesn't exist
inside the container.

Changing to bash as some of the bash constructs are being used. The
current container image's sh is just a symlink to bash.

If core_pattern is set to pipe to some binary like
`|/bin/systemd-coredump …`, we don't do anything.

Scenarios tested:

- Installed the chart and verified if pods are in running state, (most
  of the testing has been done on the original commit).

Ref: https://github.com/yugabyte/yugabyte-db/issues/4385#issuecomment-717656848

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -227,10 +227,20 @@ spec:
           postStart:
             exec:
               command:
-                - "sh"
+                - "bash"
                 - "-c"
-                - >
+                - |
                   mkdir -p /mnt/disk0/cores;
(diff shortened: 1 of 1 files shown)


ajcaldera1 (approved):
LGTM

PR #123, opened 2022-03-14
[PLAT-2031] Add rootCA checksum to yb-masters and yb-tservers statefulsets

hkandala:
This diff is a prerequisite for k8s cert rotation changes.
Added rootCA checksum to the yb-masters and yb-tservers statefulsets
So that pods are restarted when we introduce any change in the rootCA

Test Plan:
Tested locally along with k8s cert rotation changes.

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -161,16 +161,22 @@ spec:
   template:
     metadata:
       {{- if eq .name "yb-masters" }}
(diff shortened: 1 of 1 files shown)


Arnav15 (approved):

Arnav15:
Added @bhavin192 also to review

bhavin192 (approved):
cert-manager integration should keep working with these changes. Looks good!

PR #126, opened 2022-03-16
[PLAT-520] Added kube-state-metrics support

baba230896:
### Summary

- Users can provide the custom endpoint for kube-state-metrics.
- Users can install the kube-state-metrics and the platform in the same namespace with a single flag modification.

### Testing
I have done the following testing -

1. Custom kube-state-metrics endpoint.
2. kube-state-metrics installation.
3. Platform chart will use the custom endpoint if the user provides both the installation flag and custom endpoint. I have provided a higher priority to the custom endpoint than the installation. 
4. Default platform chart installation works as intended. 

**For second commit** which has changes related to prometheus scrape configuration.

### Summary

- The kube-state-metric changed the following CPU metric name. And the Yugabyte Platform java code still uses the old metric name to show the CPU metric on Platform UI. At this place, we had two ways to solve this. Either update the java code or handle it in the Prometheus scrape configuration. ~~We have chosen the second approach and fixed the Prometheus scrape configuration.~~

- I have updated the Platform code to use the new CPU metric (kube_pod_container_resource_requests) and added the relabel in Prometheus to convert the old CPU metric (kube_pod_container_resource_requests_cpu_cores) to new. 

### Test 

- Deployed the platform using without this change and created the universe. We didn't have CPU metrics in the Platform UI at this time. Once I modified the Prometheus config using the helm upgrade, it started showing the CPU metrics.

image
image


stable/yugaware/templates/configs.yaml
```diff
@@ -347,6 +347,10 @@ data:
             regex: "(.*)"
             target_label: "container_name"
             replacement: "$1"
+          - source_labels: ["__name__", "unit"]
```
anmalysh-yb: Am I understand correctly that the recent version of kube-state-metrics is having kube_pod_container_resource_requests with core label in it, instead of the old  kube_pod_container_resource_requests_cpu_cores?

If so - why we're trying to convert to the older version instead of converting to the newer version and having recent metric name in Platform code? I think we should prefer recent version of metric names, where possible.

anmalysh-yb (approved):
LGTM.
One question.

stable/yugaware/templates/configs.yaml
baba230896: @anmalysh-yb 
> Am I understand correctly that the recent version of kube-state-metrics is having kube_pod_container_resource_requests with core label in it, instead of the old kube_pod_container_resource_requests_cpu_cores?

Correct.

> If so - why we're trying to convert to the older version instead of converting to the newer version and having recent metric name in Platform code? I think we should prefer recent version of metric names, where possible.

We're doing similar for few node exporter metrics.

anmalysh-yb: For node exporter metrics - we're converting older node-exporter version metrics names to the latest version metric names.
We should do the same for kube metrics, instead of converting recent version metric names to the older version.

baba230896: @anmalysh-yb I tried the suggested approach, and it's a fairly straightforward change. The only concern I have is we won't be able to see historical data because moving forward, it'll use the new metric name, and the Prometheus have historical data for the old metric. Could you please let me know your opinion on this?

anmalysh-yb: I think that's acceptable to loose historical view in Platform UI in case we do up-versions like that.
We still have data in Prometheus UI in case we need it for some troubleshooting.

baba230896:
@anmalysh-yb I have modified the changes as per the suggestion. Could you please review the latest change and this related [PR](https://github.com/yugabyte/yugabyte-db/pull/12100)?

baba230896:
@anmalysh-yb Please check it once.

anmalysh-yb (approved):
Nice!

stable/yugaware/templates/configs.yaml
```diff
@@ -347,6 +347,14 @@ data:
             regex: "(.*)"
             target_label: "container_name"
             replacement: "$1"
+          - source_labels: ["__name__"]
```
bhavin192, line 352:
```suggestion
          # rename old name of the CPU metric to the new name and label
          # ref: https://github.com/kubernetes/kube-state-metrics/blob/master/CHANGELOG.md#v200-alpha--2020-09-16
          - source_labels: ["__name__"]
```

stable/yugaware/templates/kube-state-metrics-clusterrole.yaml
```diff
@@ -0,0 +1,179 @@
+{{- if and .Values.kubeStateMetric.install (not .Values.kubeStateMetric.customEndpoint) -}}
```
bhavin192:
```suggestion
# Adapted from https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics (Apache-2.0 license)
{{- if and .Values.kubeStateMetric.install (not .Values.kubeStateMetric.customEndpoint) -}}
```

stable/yugaware/templates/kube-state-metrics-deployment.yaml
```diff
@@ -0,0 +1,66 @@
+{{- if and .Values.kubeStateMetric.install (not .Values.kubeStateMetric.customEndpoint) -}}
```
bhavin192:
```suggestion
# Adapted from https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics (Apache-2.0 license)
{{- if and .Values.kubeStateMetric.install (not .Values.kubeStateMetric.customEndpoint) -}}
```

stable/yugaware/templates/kube-state-metrics-deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}-ksm
+  {{- if .Values.yugaware.service.annotations }}
+  annotations:
+{{ toYaml .Values.yugaware.service.annotations | indent 4 }}
+  {{- end }}
```
bhavin192: These are yugaware specific annotations, so we should remove this I guess?

stable/yugaware/templates/kube-state-metrics-deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+      app: {{ .Release.Name }}-yugaware-ksm
+  replicas: {{ .Values.kubeStateMetric.replicas }}
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}-yugaware-ksm
+      annotations:
+{{- if .Values.yugaware.pod.annotations }}
+{{ toYaml .Values.yugaware.pod.annotations | indent 8 }}
+{{- end }}
```
bhavin192: Same here. These are yugaware specific annotations, so we should remove this I guess?

stable/yugaware/templates/kube-state-metrics-service-account.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  labels:
+    app: {{ .Release.Name }}-yugaware-ksm
+    chart: {{ template "yugaware.chart" . }}
+    release: {{ .Release.Name }}
+  name: {{ .Release.Name }}-ksm
+{{- if .Values.yugaware.serviceAccountAnnotations }}
+  annotations:
+{{ toYaml .Values.yugaware.serviceAccountAnnotations | indent 4 }}
+{{- end }}
```
bhavin192: Here as well. These are annotations for yugaware's service account, let's not reuse those. We can add something like `.Values.kubeStateMetrics.serviceAccountAnnotations` later if there is a need.

stable/yugaware/templates/kube-state-metrics-service.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: {{ .Release.Name }}-ksm
+  labels:
+    app: {{ .Release.Name }}-yugaware-ksm
+    chart: {{ template "yugaware.chart" . }}
+    release: {{ .Release.Name }}
+  {{- if .Values.yugaware.service.annotations }}
+  annotations:
+  {{ toYaml .Values.yugaware.service.annotations | indent 4 }}
+  {{- end }}
```
bhavin192: Can remove this as well.

stable/yugaware/values.yaml
```diff
@@ -176,3 +176,36 @@ additionalAppConf:
...
+      memory: 64Mi
+    requests:
+      cpu: 10m
+      memory: 32Mi
+  ## List of additional cli arguments to configure kube-state-metrics
+  ## for example: --enable-gzip-encoding, --log-file, etc.
+  ## all the possible args can be found here: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/cli-arguments.md
+  extraArgs: []
+  ## As we know platform need only nodes and pods metrics from kube-state-metrics
+  ## So we just enable the collector for nodes and pods.
```
```suggestion
  ## As we know platform need only pods metrics from kube-state-metrics
  ## So we just enable the collector for pods.
```

stable/yugaware/values.yaml
```diff
@@ -176,3 +176,36 @@ additionalAppConf:
 ## to modify this unless you are using helm template command i.e. GKE
 ## app's deployer image against a Kubernetes cluster >= 1.21.
 # pdbPolicyVersionOverride: "v1beta1"
+
+## kube-state-metric support
+## Custom endpoint has higher priority than installation. 
+kubeStateMetric:
```
```suggestion
## kube-state-metrics support
## Custom endpoint has higher priority than installation. 
kubeStateMetrics:
```
To make it consistent with the project name. This will need a change at all the places where it is used.

stable/yugaware/templates/_helpers.tpl
```diff
@@ -112,3 +112,16 @@ Make list of allowed CORS origins
 {{- end -}}
 ]
 {{- end -}}
+
+{{/*
+kube-state-metrics
+*/}}
+{{- define "kubeStateMetric.endpoint" -}}
```
bhavin192:
```suggestion
{{- define "yugaware.kubeStateMetrics.endpoint" -}}
```

> A popular naming convention is to prefix each defined template with the name of the chart: {{ define "mychart.labels" }}.

https://helm.sh/docs/chart_template_guide/named_templates/#declaring-and-using-templates-with-define-and-template

stable/yugaware/values.yaml
```diff
@@ -176,3 +176,36 @@ additionalAppConf:
 ## to modify this unless you are using helm template command i.e. GKE
 ## app's deployer image against a Kubernetes cluster >= 1.21.
 # pdbPolicyVersionOverride: "v1beta1"
+
+## kube-state-metric support
+## Custom endpoint has higher priority than installation.
```
```suggestion
## Custom endpoint has higher priority than installation. The endpoint value defaults to kube-state-metrics.kube-system.svc.<domainName>:8080.
```

bhavin192 (changes requested):
Overall this looks good, I have suggested few changes.

monotek:
This could be likely done via chart dependency to the official kube-state-metrics helm chart: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics

PR #127, opened 2022-03-21
[Platform] Added HTTP to HTTPS redirection in nginx conf

baba230896:
### Summary
 
- We have made these changes because it redirects to the HTTP after integration in a few of the SSO or OIDC integration.
And previously, we disabled the HTTP traffic totally once we enabled the TLS. But after these changes, it'll redirect the HTTP traffic to HTTPS in case TLS enabled deployment.

### Test
- Tested the HTTP redirection to HTTPS. 
- Tested the upgrade of TLS enabled installation,

### DIFF
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -156,6 +156,7 @@ metadata:
 data:
   default.conf: |
 {{- if .Values.tls.enabled }}
+    # Ref: https://www.nginx.com/blog/http-strict-transport-security-hsts-and-nginx/
     server {
       listen  {{ eq .Values.ip_version_support "v6_only" | ternary "[::]:8080" "8080" }};
       server_name {{ .Values.tls.hostname }};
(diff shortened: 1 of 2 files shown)


bhavin192 (approved):
LGTM.

~~- Update commit message / PR body with why we are doing this.~~
~~- Test upgrade of TLS enabled installation, ideally changes in the service should reflect automatically.~~

sb-yb (approved):

sb-yb:
@baba230896 I am having second thoughts about this change.
Most likely our session cookie is secure ... but still I am not comfortable with this fix.
It is less secure and susceptible to man-in-the-middle attack.
Can you please add [HSTS](https://www.nginx.com/blog/http-strict-transport-security-hsts-and-nginx/) config as well ASAP?

This will ensure that browser will rewrite any http urls to https. Thus no in-flight risk. 

Even better would be setting our play framework to use trusted proxy. But that involves lot more setup and testing and I do not have bandwidth for that right now.

sb-yb:
Another option is to use proxy_redirect directive and rewrite location header to switch http redirects to https.

sb-yb:
@baba230896 
So looks like adding the missing 
```
proxy_set_header X-Forwarded-Proto $scheme;
```
directive fixed the problem. 
Now all the redirects from app are to going to https. 
I have only tested this on yugabundle. 
Can you please test this for k8s and remove the http to https forwarding?

sb-yb (approved):

sb-yb:
Thanks for making the change!

PR #128, opened 2022-03-29
[PLAT-3281] Add option to pass nodeCert and clientCert to support Hashicorp Certificates

hkandala:
Add nodeCert and clientCert in values.yaml in order to pass generated node cert or client cert.
Use the same cert for all nodes in this scenario.
This is needed as Hashicorp Vault certificates won't have private key and helm charts won't be able to generate certs.
It will default to old logic if rootCA.key is not null

Tested locally along with the platform changes done for Hashicorp Vault support


stable/yugabyte/values.yaml
```diff
@@ -71,6 +71,15 @@ tls:
...
+  # nodeCert and clientCert will be used only when rootCA.key is null
+  # Will be ignored and genSignedCert will be used to generate
+  # node and client certs if rootCA.key is provided
```
```suggestion
  ## nodeCert and clientCert will be used only when rootCA.key is empty.
  ## Will be ignored and genSignedCert will be used to generate
  ## node and client certs if rootCA.key is provided.
  ## cert and key are base64 encoded content of certificate and key.
```

Convention is to use double # for comments.

stable/yugabyte/values.yaml
```diff
@@ -71,6 +71,15 @@ tls:
...
+  nodeCert:
+    cert: null
+    key: null
+  clientCert:
+    cert: null
+    key: null
```
```suggestion
  nodeCert:
    cert: ""
    key: ""
  clientCert:
    cert: ""
    key: ""
```
Usually a blank string is used instead of null. It works the same on a blank string (do test it once though).

bhavin192 (approved):
LGTM! added two minor suggestion based on conventions I have seen with Helm charts in general.

stable/yugabyte/values.yaml
hkandala: Updated, thanks!

stable/yugabyte/values.yaml
hkandala: Done!

PR #129, opened 2022-04-04
[Plat-3409][k8s] Run yugabyte as non root user

baba230896:
### Summary

- Run Yugabyte DB as a non-root user. 

### Test

- I used the changes to deploy TLS enabled & non-TLS universes and tested them with `ysqlsh` & `ycqlsh` on K8s, and it's working fine.

- For OCP, either we need to provide the `user ID`, `fsgroup` and `group ID` as per the OCP projects configuration OR keep the charts' based security context disabled and let them (OCP) add on their own.

- Current changes won't support helm upgrade from root to non-root users.

### DIFF
--- a/stable/yugabyte/values.yaml
+++ b/stable/yugabyte/values.yaml
@@ -309,3 +309,14 @@ preflight:
   skipAll: false
   # Set to true to skip port bind checks
   skipBind: false
+
+## Pod securityContext
+## Ref: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context
+## Run YB Master and YB Tserver as a non-root user
+podSecurityContext:
+  enabled: false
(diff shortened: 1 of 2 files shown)


stable/yugabyte/values.yaml
```diff
@@ -309,3 +309,14 @@ preflight:
   skipAll: false
   # Set to true to skip port bind checks
   skipBind: false
+
+## Pod securityContext
+## Ref: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context
+## Run YB Master and YB Tserver as a non-root user
```
```suggestion
## The following configuration runs YB-Master and YB-TServer as a non-root user
```

bhavin192 (approved):
LGTM! Since it is disable by default, it won't affect existing installations.

PR #132, opened 2022-05-09
[Backport 2.12][Plat-3409][k8s] Run yugabyte as non root user (#129)

baba230896:
* [Plat-3409][k8s] Run yugabyte as non-root user

Test -
I used the changes to deploy TLS enabled & non-TLS universes and tested them with ysqlsh & ycqlsh on K8s, and it's working fine.
For OCP, either we need to provide the user ID, fsgroup and group ID as per the OCP projects configuration OR
keep the charts' based security context disabled and let them (OCP) add on their own.

Current changes won't support helm upgrade from root to non-root users.

* Update stable/yugabyte/values.yaml

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -187,6 +187,9 @@ spec:
       imagePullSecrets:
       - name: {{ $root.Values.Image.pullSecretName }}
       {{ end }}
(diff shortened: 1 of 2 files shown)


bhavin192 (approved):
Please mention the original commit id `c62d8cddfc0c713ccad87bd718733c4abd2e381f` in this commit while squash and merging.

PR #133, opened 2022-05-30
[PLAT-3560][PLAT-3950] Add ability to run platform as a non-root user

baba230896:
### Summary 
- Supports to run platform as a non-root user.
- Supports the platform upgrade from root user to a non-root user.
- Supports the platform upgrade from 1001 user to a 10001 user.
- Added the script in the init container to handle the SSH keys permission issue while restarting.

### Test plan
1. Deploy the platform using the chart changes with the security context enabled and able to spin up the K8s and GCP VM based universes successfully.
2. Generated the template by chart changes, and without chart changes and comparison look good, which means changes won't impact the regular deployment flow.
3. Deployed the platform v2.12.3.0-b19 [Chart v2.12.3] and configured the GCP cloud provider. Then upgraded it to the platform v2.13.3.0-b117 along with chart changes with security context enabled and successfully deployed the VM based universes.
4. To test the init container script - Deployed the platform using the changes and configured the GCP Cloud provider. Restarted the sts using the helm rollout command. The script worked as intended and was able to deploy VM based universe successfully.

### Summary

- Added `fsGroupPolicyChange` in charts. However, Chart will add it only on clusters above or equal to v1.20.
- The Chart will create ConfigMap for init-container-script only when we have security context enabled on the Platform. 

### Test Plan
1. Test sts rollout on 1.19 GKE cluster

- I spun up a 1.19 GKE cluster and deployed the Platform running as non-root using chart changes. Changes won't add `fsGroupPolicyChange` because v1.19 does not have that feature. 
- Configured the GCP Cloud provider and created one certificate in Security Tab. 
- Restart the STS that updates the permissions of SSH keys. However, the init-container script will take this up as a trigger and change the permissions of all SSH keys to 400 to resolve the permission issue. 
- Successfully created the VM-based universe using the previously configured provider and TLS certs & added one more node after universe creation. 

2. Upgrade test from Platfrom v2.12 [root] to v2.13 [non-root] on GKE v1.22. 

- I deployed the Platform as root using Chart v2.12.3 and Platform image 2.12.3.0-b19. 
- Configured the GCP Cloud provider and created one certificate in Security Tab. 
- I upgraded the Platform using the changes as non-root and v2.13. Chart added `fsGroupPolicyChange`, but `fsGroup` updated the permission because the directory owner doesn't match, so `fsGroupPolicyChange` won't prevent us from SSH permission issue in this case. However, the init-script captures this event and changes the permissions of all SSH keys to 400 to resolve the permission issue of SSH keys.
- Successfully created the VM-based universe using the previously configured provider and TLS certs & added one more node after universe creation. 

3. This change won't impact regular deployment because it needs `securityContext.enabled: true` to run the Platform as non-root. By default, we kept it false.

### DIFF
--- a/stable/yugaware/templates/init-container-script.yaml
+++ b/stable/yugaware/templates/init-container-script.yaml
@@ -1,0 +1,49 @@
+{{- if .Values.securityContext.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-yugaware-init-container
+  labels:
+    app: {{ template "yugaware.name" . }}
+    chart: {{ template "yugaware.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Values.helm2Legacy | ternary "Tiller" (.Release.Service | quote) }}
+data:
+  init-permissions.sh: |
+    #!/bin/bash
+
+    set -xe
+
+    directory_name="/opt/yugabyte/yugaware/data"
+    pemfiles=$(find "${directory_name}/keys/" -name "*.pem" -exec stat -c "%a" {} + | uniq | tr '\n' ',')
+    IFS="," read -r -a array <<< "$pemfiles"
+    
+    trigger=false
+    echo "Finding pem files without 400 permission"
+    
+    for pemfile in  "${array[@]}"
+    do
+      echo "pemfile - $pemfile"
+      if [[ "$pemfile" != *400* ]]; then
+        echo "pemfile - $pemfile"
+        trigger=true
+        break
+      fi
+    done
+    
+    if ${trigger}; then
+      echo "Creating copy of data/keys directory"
+      cp -r "${directory_name}/keys" "${directory_name}/new_keys"
+
+      echo "Setting permission to all pem files to 400"
+      find "${directory_name}/new_keys/" -name "*.pem" -exec chmod 400 {} +
+
+      echo "Renaming old keys directory"
+      mv "${directory_name}/keys" "${directory_name}/keys-$(date +%s)"
+
+      echo "Renaming new keys directory"
+      mv "${directory_name}/new_keys" "${directory_name}/keys"
+    else
+      echo "Does not find pem file without 400 permission"
+    fi
+{{- end }}
--- a/stable/yugaware/templates/init-container-script.yaml
+++ b/stable/yugaware/templates/init-container-script.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ .Release.Name }}-yugaware-init-container
+  name: {{ .Release.Name }}-yugaware-init
   labels:
     app: {{ template "yugaware.name" . }}
     chart: {{ template "yugaware.chart" . }}
@@ -12,20 +12,18 @@ data:
   init-permissions.sh: |
     #!/bin/bash
 
-    set -xe
+    set -xe -o pipefail
 
-    directory_name="/opt/yugabyte/yugaware/data"
-    pemfiles=$(find "${directory_name}/keys/" -name "*.pem" -exec stat -c "%a" {} + | uniq | tr '\n' ',')
-    IFS="," read -r -a array <<< "$pemfiles"
+    data_directory="/opt/yugabyte/yugaware/data"
+    pemfiles=$(find "${data_directory}/keys/" -name "*.pem" -exec stat -c "%a" {} + | uniq | tr '\n' ',')
+    IFS="," read -r -a pemfile_perms <<< "${pemfiles}"
     
     trigger=false
-    echo "Finding pem files without 400 permission"
+    echo "Finding pem files with permissions different than 400, and setting their permissions to 400."
     
-    for pemfile in  "${array[@]}"
-    do
-      echo "pemfile - $pemfile"
-      if [[ "$pemfile" != *400* ]]; then
-        echo "pemfile - $pemfile"
+    for pemfile in  "${pemfile_perms[@]}"; do
+      if [[ "${pemfile}" != *400* ]]; then
+        echo "Found a pem file with permissions ${pemfile}"
         trigger=true
         break
       fi
@@ -33,17 +31,17 @@ data:
     
     if ${trigger}; then
       echo "Creating copy of data/keys directory"
-      cp -r "${directory_name}/keys" "${directory_name}/new_keys"
+      cp -r "${data_directory}/keys" "${data_directory}/new_keys"
 
-      echo "Setting permission to all pem files to 400"
-      find "${directory_name}/new_keys/" -name "*.pem" -exec chmod 400 {} +
+      echo "Setting permission of all pem files to 400"
+      find "${data_directory}/new_keys/" -name "*.pem" -exec chmod 400 {} +
 
-      echo "Renaming old keys directory"
-      mv "${directory_name}/keys" "${directory_name}/keys-$(date +%s)"
+      echo "Renaming existing keys directory"
+      mv "${data_directory}/keys" "${data_directory}/keys-$(date +%s)"
 
       echo "Renaming new keys directory"
-      mv "${directory_name}/new_keys" "${directory_name}/keys"
+      mv "${data_directory}/new_keys" "${data_directory}/keys"
     else
-      echo "Does not find pem file without 400 permission"
+      echo "All pem files already have permission set to 400"
     fi
 {{- end }}
--- a/stable/yugaware/templates/statefulset.yaml
+++ b/stable/yugaware/templates/statefulset.yaml
@@ -82,10 +82,30 @@ spec:
         - image: {{ include "full_yugaware_image" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           name: prometheus-configuration
-          command: ["cp", "/default_prometheus_config/prometheus.yml", "/prometheus_configs/prometheus.yml"]
+          command:
+            - 'bash'
+            - '-c'
+            - >
+              cp /default_prometheus_config/prometheus.yml /prometheus_configs/prometheus.yml;
+              echo "finding pem files with 660 permission";
+              directory_name="/opt/yugabyte/yugaware/data";
+              pemfiles=$(find "$directory_name"/keys/** -name "*.pem" -exec stat -c "%a" {} +);
+              if [[ $pemfiles == *660* ]]; then
+                echo "creating new key directory";
+                mkdir "$directory_name"/new_keys;
+                echo "copying content of data/key/* to new keys directory";
+                cp -r "$directory_name"/keys/* "$directory_name"/new_keys/;
+                echo "setting permission to all pem files to 600";
+                find "$directory_name"/new_keys/** -name "*.pem" -exec chmod 600 {} +;
+                echo "Renaming old keys directory";
+                mv "$directory_name"/keys "$directory_name"/keys-"$(date +%s)";
+                echo "Renaming new keys directory";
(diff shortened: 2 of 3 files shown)


stable/yugaware/templates/statefulset.yaml
```diff
@@ -98,6 +96,11 @@ spec:
         - name: postgres
           image: {{ include "full_image" (dict "containerName" "postgres" "root" .) }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.securityContext.enabled }}
+          securityContext:
+            runAsUser: 999
```
srinivasa-vasu: I understand the Postgres user in the container image is with the UID `999`. This might break the upgrade path if someone is already running as a non-root with a different UID. Instead of hard-coding it, it's better to externalize in the values.yaml.

stable/yugaware/values.yaml
```diff
@@ -114,8 +114,14 @@ tls:
...
-  fsGroup: 1001
-  runAsUser: 1001
+
+  podLevel:
+    fsGroup: 10001
+
+  containerLevel:
+    runAsUser: 10001
+    runAsGroup: 10001
+    runAsNonRoot: true
```
bhavin192: This won't give the user much flexibility, and is confusing. Because we will be setting same value for all the containers when we set `containerLevel` and conceptually, podLevel one is supposed to do the same thing i.e. change the securityContext for all the container (though it has extra options like fsGroup). 

How about something like this.
```yaml
securityContext:
  enabled: false
  runAsUser: 10001
  runAsGroup: 10001
  fsGroup: 10001
```

And then, inside every section like `yugaware`, `prometheus`, `nginx`, `postgres` we will have `containerSecurityContext`. This will be blank `{}` for now, i.e. we won't set it explicitly. 

For upgrade scenarios root -> non-root, we will have a comment near `securityContext`, if you are upgrading from root -> non-root setup, make sure you set `postgres.containerSecurityContext.runAsUser` to `999`.

stable/yugaware/templates/statefulset.yaml
```diff
@@ -134,9 +137,13 @@ spec:
           image: {{ include "full_image" (dict "containerName" "prometheus" "root" .) }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           {{- if (not .Values.ocpCompatibility.enabled) }}
+          {{- if .Values.securityContext.enabled }}
+          securityContext: {{- .Values.securityContext.containerLevel | toYaml | nindent 12 }}
+          {{- else }}
           securityContext:
             runAsUser: 0
           {{- end }}
+          {{- end }}
```
bhavin192:
```suggestion
          {{- if .Values.securityContext.enabled }}
          securityContext: {{- .Values.securityContext.containerLevel | toYaml | nindent 12 }}
          {{- else }}
          {{- if (not .Values.ocpCompatibility.enabled) }}
          securityContext:
            runAsUser: 0
          {{- end }}
          {{- end }}
```

With this, we still allow one to set own securityContext when running on OCP as well.

bhavin192 (changes requested):

stable/yugaware/values.yaml
bhavin192: Based on offline discussion:
We will have the values I suggested above.
And to make sure we have backward compatibility, we will have following settings by default:
```yaml
postgres:
  containerSecurityContext:
    enabled: true
    runAsUser: 999
```
That way even if someone had `securityContext.enabled=true`, their setup will keep working, because previously their postgres will be running as postgres user (999). 
And we can add a note on `securityContext`: if you would like to have same UID across all the containers then make sure to set `postgres.containerSecurity.enabled=false` while installing the chart.

stable/yugaware/templates/statefulset.yaml
baba230896: It won't happen unless they have modified the platform charts according to their needs. Because as per old changes, the platform charts do not change the user for the Postgres container. However, as per new changes, it will change the user for the Postgres container if we want. Otherwise, it'll use 999 as default.

stable/yugaware/values.yaml
baba230896: I modified the changes as per the offline discussion.

stable/yugaware/templates/statefulset.yaml
```diff
@@ -82,17 +82,40 @@ spec:
         - image: {{ include "full_yugaware_image" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           name: prometheus-configuration
-          command: ["cp", "/default_prometheus_config/prometheus.yml", "/prometheus_configs/prometheus.yml"]
+          command:
+            - 'bash'
+            - '-c'
+            - >
+              cp /default_prometheus_config/prometheus.yml /prometheus_configs/prometheus.yml;
+              echo "finding pem files with 660 permission";
```
iSignal: minor: let's move this to a standalone script file and 
```
set -euxo pipefail
```
on it so that we can capture the work done by the script in kubectl logs and fail on any errors.

stable/yugaware/templates/statefulset.yaml
```diff
@@ -72,7 +75,14 @@ spec:
             items:
               - key: prometheus.yml
                 path: prometheus.yml
-
+        {{- if .Values.securityContext.enabled }}
+        - name: init-container-script
+          configMap:
+            name: {{ .Release.Name }}-yugaware-init-container
```
bhavin192:
```suggestion
            name: {{ .Release.Name }}-yugaware-init
```
Mostly to use less number of characters.

stable/yugaware/values.yaml
```diff
@@ -114,8 +114,11 @@ tls:
 ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
 securityContext:
   enabled: false
-  fsGroup: 1001
-  runAsUser: 1001
+  fsGroup: 10001
+  runAsUser: 10001
+  runAsGroup: 10001
+  runAsNonRoot: true
+  fsGroupChangePolicy: "OnRootMismatch"
```
```suggestion
  ## fsGroup related values are set at the pod level.
  fsGroup: 10001
  fsGroupChangePolicy: "OnRootMismatch"
  ## The following values are set for yugaware and prometheus
  ## containers.
  runAsUser: 10001
  runAsGroup: 10001
  runAsNonRoot: true
```
This will change later, but for now it represents current state.

stable/yugaware/templates/init-container-script.yaml
```diff
@@ -0,0 +1,49 @@
...
+  labels:
+    app: {{ template "yugaware.name" . }}
+    chart: {{ template "yugaware.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Values.helm2Legacy | ternary "Tiller" (.Release.Service | quote) }}
+data:
+  init-permissions.sh: |
+    #!/bin/bash
+
+    set -xe
```
bhavin192:
```suggestion
    set -xe -o pipefail
```
This would be good to have.

stable/yugaware/templates/init-container-script.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+    directory_name="/opt/yugabyte/yugaware/data"
```
bhavin192:
```suggestion
    data_directory="/opt/yugabyte/yugaware/data"
```

stable/yugaware/templates/init-container-script.yaml
```diff
@@ -0,0 +1,49 @@
...
+    pemfiles=$(find "${directory_name}/keys/" -name "*.pem" -exec stat -c "%a" {} + | uniq | tr '\n' ',')
+    IFS="," read -r -a array <<< "$pemfiles"
```
bhavin192:
```suggestion
    IFS="," read -r -a pemfile_perms <<< "$pemfiles"
```
Couldn't think of anything better, but should be readable than just array.

stable/yugaware/templates/init-container-script.yaml
```diff
@@ -0,0 +1,49 @@
...
+    
+    trigger=false
+    echo "Finding pem files without 400 permission"
```
bhavin192:
```suggestion
    echo "Finding pem files with permissions different than 400, and setting their permissions to 400."
```

stable/yugaware/templates/init-container-script.yaml
```diff
@@ -0,0 +1,49 @@
...
+    
+    for pemfile in  "${array[@]}"
+    do
+      echo "pemfile - $pemfile"
+      if [[ "$pemfile" != *400* ]]; then
+        echo "pemfile - $pemfile"
+        trigger=true
+        break
```
bhavin192:
```suggestion
      if [[ "$pemfile" != *400* ]]; then
        echo "Found a pem file with permissions ${pemfile}"
        trigger=true
        break
```

stable/yugaware/templates/init-container-script.yaml
```diff
@@ -0,0 +1,49 @@
...
+      fi
+    done
+    
+    if ${trigger}; then
+      echo "Creating copy of data/keys directory"
+      cp -r "${directory_name}/keys" "${directory_name}/new_keys"
+
+      echo "Setting permission to all pem files to 400"
```
bhavin192:
```suggestion
      echo "Setting permission of all pem files to 400"
```

stable/yugaware/templates/init-container-script.yaml
```diff
@@ -0,0 +1,49 @@
...
+      find "${directory_name}/new_keys/" -name "*.pem" -exec chmod 400 {} +
+
+      echo "Renaming old keys directory"
```
bhavin192:
```suggestion
      echo "Renaming existing keys directory"
```

stable/yugaware/templates/init-container-script.yaml
```diff
@@ -0,0 +1,49 @@
...
+      mv "${directory_name}/keys" "${directory_name}/keys-$(date +%s)"
+
+      echo "Renaming new keys directory"
+      mv "${directory_name}/new_keys" "${directory_name}/keys"
+    else
+      echo "Does not find pem file without 400 permission"
```
bhavin192:
```suggestion
      echo "All pem files already have permission set to 400"
```

bhavin192 (approved):
This LGTM now :tada:

stable/yugaware/templates/statefulset.yaml
```diff
@@ -82,17 +92,33 @@ spec:
...
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           name: prometheus-configuration
-          command: ["cp", "/default_prometheus_config/prometheus.yml", "/prometheus_configs/prometheus.yml"]
           {{- if .Values.securityContext.enabled }}
+          command:
+            - 'bash'
+            - '-c'
+            - >
+              cp /default_prometheus_config/prometheus.yml /prometheus_configs/prometheus.yml;
+              bash /init-container/init-permissions.sh;
```
iSignal, line 101: if this script fails, the error may not be propagated to the bash -c launcher unless we either set -e on this small piece OR run it as `cp ... && bash init-permissions.sh`

stable/yugaware/values.yaml
```diff
@@ -114,8 +114,14 @@ tls:
...
 securityContext:
   enabled: false
-  fsGroup: 1001
-  runAsUser: 1001
+  ## fsGroup related values are set at the pod level.
+  fsGroup: 10001
+  fsGroupChangePolicy: "OnRootMismatch"
+  ## The following values are set for yugaware and prometheus
+  ## containers.
+  runAsUser: 10001
```
iSignal, line 122: minor: add a comment saying that if runAsUser is set to a diff value other than 10001, it will cause VM universe workflows to fail.

PR #136, opened 2022-08-25
PLAT-5217 YBC changes to add releases volume mount

saratkumar-yb:
PLAT-5217 YBC changes to add releases volume mount

### DIFF
--- a/stable/yugaware/templates/statefulset.yaml
+++ b/stable/yugaware/templates/statefulset.yaml
@@ -256,6 +256,9 @@ spec:
           - name: yugaware-storage
             mountPath: /opt/yugabyte/releases/
             subPath: releases
+          - name: ybc-storage
+            mountPath: /opt/yugabyte/ybc/releases/
+            subPath: ybc_releases
           # old path for backward compatibility
           - name: yugaware-storage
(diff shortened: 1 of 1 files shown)


stable/yugaware/templates/statefulset.yaml
```diff
@@ -256,6 +256,9 @@ spec:
           - name: yugaware-storage
             mountPath: /opt/yugabyte/releases/
             subPath: releases
+          - name: ybc-storage
```
skorobogatydmitry, line 259: There should be volume named this way. See line 40.

stable/yugaware/templates/statefulset.yaml
```diff
@@ -256,6 +256,9 @@ spec:
...
+            mountPath: /opt/yugabyte/ybc/releases/
```
skorobogatydmitry, line 260: What is it for ? Do we have HA story for this data ?
There should be a way to backup & restore it. Ideally, would be nice to be able to remove it completely without any harm to the Platform.

@iSignal, can you take a look as well ?

skorobogatydmitry (changes requested):

PR #137, opened 2022-09-05
Fix tolerations

leovigna:


leovigna:
Fix toleratiosn & bump version.
Have some linting issues at the moment that I could fix if there is interest in this PR.

leovigna:
outdated

PR #149, opened 2023-01-06
Update find_docker_tag.py

govardhanjalla:
Switch to v2 of the docker tag registry, v1 seems to be gone.


iSignal (approved):

bhavin192 (approved):
We have this in master. LGTM.

PR #152, opened 2023-01-13
[PLAT-5675][Platform] Map extra k8s secret to yb db pods in Platform-created namespace.

govardhanjalla:
Summary:
We create secrets from other namespaces if required for master.secretEnv and tserver.secretEnv. We mangle names to make unique secret names by following the [doc](https://docs.google.com/document/d/1rGoy2pTaUXI7NfjDCi02Cdq1UeJoCyQEKy6uykKqwlo/edit?pli=1# ).
If two same secrets are mentioned in both master/tserver secretEnv we create two separate secrets.

Only helm chart changes are needed for this feature.

Helm chart installation fails if namespace/secret/key doesn't exist (if optional is not set) or if permissions are not there to look up secrets from different namespace.

If same secret is specified in master.secretEnv (or tserver.secretEnv) more than once, installation fails. We are not catching this as of now.

Test Plan:
Manually tested following scenarios.
In values.yaml I gave
```
  secretEnv:
  - name: envusername
    valueFrom:
      secretKeyRef:
        name: secret-basic-auth
        key: username
        namespace: yb-admin-gjalla-test
        optional: true
```
 and ran helm install and saw the secret "yw-yb-ad-secre-usern-23a9-master" getting created in universe namespace. Tried out error cases like
Specified a secret that doesn't exist with optional set to false - fails with message "Secret or key missing for secret-basic-auth/user_name in namespace: yb-admin-gjalla-test"
master.secretEnv and tserver.secretEnv having same secrets - installation succeeds as we create two different secrets (they differ in suffix names)
master.secretEnv having same entry twice - fails with "secrets "yw-yb-ad-secre-usern-23a9-master" already exists"

Reviewers: sanketh, nsingh, bgandhi

Subscribers: yugaware

### DIFF
--- a/stable/yugabyte/templates/_helpers.tpl
+++ b/stable/yugabyte/templates/_helpers.tpl
@@ -57,6 +57,89 @@ release: {{ .root.Release.Name | quote }}
 {{- end }}
 
 {{/*
+Create secrets in DBNamespace from other namespaces by iterating over envSecrets.
+*/}}
+{{- define "yugabyte.envsecrets" -}}
+{{- range $v := .secretenv }}
+{{- if $v.valueFrom.secretKeyRef.namespace }}
+{{- $secretObj := (lookup
+"v1"
+"Secret"
+$v.valueFrom.secretKeyRef.namespace
+$v.valueFrom.secretKeyRef.name)
+| default dict }}
+{{- $secretData := (get $secretObj "data") | default dict }}
+{{- $secretValue := (get $secretData $v.valueFrom.secretKeyRef.key) | default "" }}
+{{- if (and (not $secretValue) (not $v.valueFrom.secretKeyRef.optional)) }}
+{{- required (printf "Secret or key missing for %s/%s in namespace: %s"
+$v.valueFrom.secretKeyRef.name
+$v.valueFrom.secretKeyRef.key
+$v.valueFrom.secretKeyRef.namespace)
+nil }}
+{{- end }}
+{{- if $secretValue }}
+apiVersion: v1
+kind: Secret
+metadata:
+  {{- $secretfullname := printf "%s-%s-%s-%s"
+  $.root.Release.Name
+  $v.valueFrom.secretKeyRef.namespace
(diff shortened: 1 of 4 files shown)


stable/yugabyte/templates/service.yaml
```diff
@@ -24,7 +24,13 @@ data:
...
 ---
 {{- end }}
-
+--- # master secret env
+{{- $data := dict "secretenv" $.Values.master.secretEnv "root" . "suffix" "master"}}
+{{- include "yugabyte.envsecrets" $data }}
+--- # tserver secret env
+{{- $data := dict "secretenv" $.Values.tserver.secretEnv "root" . "suffix" "tserver" }}
+{{- include "yugabyte.envsecrets" $data }}
+---
```
bhavin192: minor: this can be in a separate file as well. I personally don't have any strong opinion on whether it should be here or in a separate file. 
Maybe our comment here should say something like env secrets from other namespaces to be more clear?

stable/yugabyte/templates/_helpers.tpl
```diff
@@ -56,6 +56,89 @@ release: {{ .root.Release.Name | quote }}
 {{- end }}
 {{- end }}
 
+{{/*
+Generate env secrets
```
bhavin192: I think it is worth clarifying a bit more on what we are doing here, i.e. copying secrets from other namespace and creating those in the DB namespace.

bhavin192 (approved):
Looks good overall! 
The idea of copying secrets seems a kind of hack, but I see why we are doing it. And can't think of any other way as of now. 

Since this is change in values.yaml, it would be better to add an example with some details there. It will serve as documentation for new user facing options in the chart i.e. the namespace field.

govardhanjalla:
> Looks good overall! The idea of copying secrets seems a kind of hack, but I see why we are doing it. And can't think of any other way as of now.
> 
> Since this is change in values.yaml, it would be better to add an example with some details there. It will serve as documentation for new user facing options in the chart i.e. the namespace field.

Added comments to existing example.

PR #153, opened 2023-01-16
[PLAT-5753] Added metric name relabeling for CPU metric

baba230896:
## Summary 
- Modified the config to keep new and old CPU metrics.
- Replaced the new metric name with the old one.

### Test - 1
- Spun up the EKS and deployed the platform using the `metric_relabel_config` change. Created the universe in the same namespace.
- Deployed the kube-state-metric with the latest version (2.7.0) and checked the CPU metrics on YBA UI.
- Having expected CPU metrics.

### Test - 2
- Used the following `bitnami/kube-state-metrics:1.9.7` image to verify the change with the older version. 
- Having expected CPU metrics.

fixes https://yugabyte.atlassian.net/browse/PLAT-5753

### DIFF
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -441,7 +441,7 @@ data:
             target_label: "__name__"
             replacement: "kube_pod_container_resource_requests_cpu_cores"
           # Keep metrics for CPU, discard duplicate metrics
-          - source_labels: ["saved_name"]
+          - source_labels: ["__name__"]
             regex: "kube_pod_container_resource_requests_cpu_cores"
             action: keep
 
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -164,7 +164,7 @@ data:
       listen  {{ eq .Values.ip_version_support "v6_only" | ternary "[::]:8080" "8080" }};
       server_name {{ .Values.tls.hostname }};
       return  301  https://$host$request_uri;
-    }  
+    }
 {{- end }}
 
     server {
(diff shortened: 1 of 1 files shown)


stable/yugaware/templates/configs.yaml
```diff
@@ -411,11 +411,11 @@ data:
 
       - job_name: 'kube-state-metrics'
         static_configs:
-        - targets: ['kube-state-metrics.kube-system.svc.{{.Values.domainName}}:8080']
+        - targets: ['kube-state-metrics.kube-state.svc.{{.Values.domainName}}:8080']
```
iSignal: our docs are asking users to install it in kube-system - https://docs.yugabyte.com/preview/yugabyte-platform/install-yugabyte-platform/prepare-environment/kubernetes/#install-kube-state-metrics 

Let's keep it consistent with that unless there's a reason to use a different namespace.

iSignal (approved):

stable/yugaware/templates/configs.yaml
baba230896: Ahh. I did it for testing. I installed multiple versions of Kube-state-metrics for testing. I forget to revert it. Will update. Thanks for capturing it.

stable/yugaware/templates/configs.yaml
```diff
@@ -434,6 +434,16 @@ data:
...
             regex: "(.*)yb-(.*)"
             action: keep
+          # rename new name of the CPU metric to the old name and label
+          # ref: https://github.com/kubernetes/kube-state-metrics/blob/master/CHANGELOG.md#v200-alpha--2020-09-16
+          - source_labels: ["__name__", "unit"]
+            regex: "kube_pod_container_resource_requests;core"
+            target_label: "__name__"
+            replacement: "kube_pod_container_resource_requests_cpu_cores"
+          # Keep metrics for CPU, discard duplicate metrics
+          - source_labels: ["saved_name"]
```
bhavin192:
```suggestion
          - source_labels: ["__name__"]
```
If we use saved_name, we don't get any metrics when using latest kube-state-metrics, because the saved_name is created at the first, and it holds the new name.

bhavin192 (approved):
Need to fix the one comment, otherwise LGTM.

PR #164, opened 2023-05-01
[PLAT-9120] [yugabyte] Decouple istio and servicePerPod

shinji62:
Decoupling service per pod and Istio
This allow to create service per pod without adding the side car injection, for example if I don't want my DB to be on my service mesh but I need to the service per pod because I would like to use XCluster without the need of pod-to-pod communication and without MCS.

Or in case of YBA been in a different namespace than YB and cross-namespace communication is not allow.

### DIFF
--- a/stable/yugabyte/templates/_helpers.tpl
+++ b/stable/yugabyte/templates/_helpers.tpl
@@ -212,7 +212,7 @@ Generate server FQDN.
     {{ $membershipName := required "A valid membership name is required! Please set multicluster.kubernetesClusterId" .Values.multicluster.kubernetesClusterId }}
     {{- printf "$(HOSTNAME).%s.%s.$(NAMESPACE).svc.clusterset.local" $membershipName .Service.name -}}
   {{- else if .Values.oldNamingStyle -}}
-    {{- printf "$(HOSTNAME).%s.$(NAMESPACE).svc.%s" .Service.name .Values.domainName -}}  
+    {{- printf "$(HOSTNAME).%s.$(NAMESPACE).svc.%s" .Service.name .Values.domainName -}}
   {{- else -}}
     {{- if .Values.multicluster.createServiceExports -}}
       {{ $membershipName := required "A valid membership name is required! Please set multicluster.kubernetesClusterId" .Values.multicluster.kubernetesClusterId }}
@@ -267,7 +267,7 @@ Generate server web interface.
 Generate server CQL proxy bind address.
 */}}
 {{- define "yugabyte.cql_proxy_bind_address" -}}
-  {{- if or .Values.istioCompatibility.enabled .Values.multicluster.createServiceExports -}}
+  {{- if or .Values.istioCompatibility.enabled .Values.multicluster.createServiceExports .Values.multicluster.createServicePerPod -}}
     0.0.0.0:{{ index .Service.ports "tcp-yql-port" -}}
   {{- else -}}
     {{- include "yugabyte.server_fqdn" . -}}
--- a/stable/yugabyte/templates/_helpers.tpl
+++ b/stable/yugabyte/templates/_helpers.tpl
@@ -206,7 +206,7 @@ Get files from fs data directories for readiness / liveness probes.
 Generate server FQDN.
 */}}
 {{- define "yugabyte.server_fqdn" -}}
-  {{- if (and .Values.istioCompatibility.enabled .Values.multicluster.createServicePerPod) -}}
+  {{- if (or .Values.istioCompatibility.enabled .Values.multicluster.createServicePerPod) -}}
     {{- printf "$(HOSTNAME).$(NAMESPACE).svc.%s" .Values.domainName -}}
   {{- else if (and .Values.oldNamingStyle .Values.multicluster.createServiceExports) -}}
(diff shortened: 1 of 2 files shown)


shinji62:
cc @iSignal

iSignal:
@shinji62 : servicePerPod can be technically used without Istio - what is the environment in which you are using this? 
@bhavin192 : could you review this?

stable/yugabyte/templates/_helpers.tpl
```diff
@@ -206,7 +206,7 @@ Get files from fs data directories for readiness / liveness probes.
 Generate server FQDN.
 */}}
 {{- define "yugabyte.server_fqdn" -}}
-  {{- if (and .Values.istioCompatibility.enabled .Values.multicluster.createServicePerPod) -}}
+  {{- if (or .Values.istioCompatibility.enabled .Values.multicluster.createServicePerPod) -}}
```
bhavin192: For all these cases you can just drop the `.Values.istioCompatibility.enabled` from this condition, because all are more related to `createServicePerPod` actually.

```yaml
  {{- if .Values.multicluster.createServicePerPod -}}
```

bhavin192 (changes requested):
The change looks good, have only one small suggestion.

shinji62:
@bhavin192 Comment addressed.
@iSignal K8S based on F5 Volterra distribution.

stable/yugabyte/templates/service.yaml
```diff
@@ -386,7 +386,7 @@ spec:
           {{- if $rpcPreflight -}}{{ $rpcPreflight | nindent 12 }}{{ end -}}
           {{- $broadcastAddr := include "yugabyte.server_broadcast_address" $serviceValues -}}
           {{/* skip bind check for Istio multi-cluster, we cannot/don't bind to service IP */}}
-          {{- if (not (and $root.Values.istioCompatibility.enabled $root.Values.multicluster.createServicePerPod)) }}
+          {{- if (not (or $root.Values.istioCompatibility.enabled $root.Values.multicluster.createServicePerPod)) }}
```
bhavin192: We can modify the comment as we are decoupling servicePerPod and istioCompatibility. 
```
{{/* skip bind check for servicePerPod multi-cluster, we cannot/don't bind to service IP */}}
```
Also, this skip thing is specific to servicePerPod so if can be just:
```
{{- if not $root.Values.multicluster.createServicePerPod }}
```

stable/yugabyte/templates/_helpers.tpl
```diff
@@ -267,7 +267,7 @@ Generate server web interface.
 Generate server CQL proxy bind address.
 */}}
 {{- define "yugabyte.cql_proxy_bind_address" -}}
-  {{- if or .Values.istioCompatibility.enabled .Values.multicluster.createServiceExports -}}
+  {{- if .Values.multicluster.createServiceExports -}}
```
bhavin192: We should not change this because when you use this chart with Istio you need the bind_address to be `0.0.0.0`. 
So, this or condition should remain as is.

bhavin192 (commented):
Only two fixes, and then we can merge this PR.

stable/yugabyte/templates/service.yaml
shinji62: done

stable/yugabyte/templates/_helpers.tpl
shinji62: done

shinji62:
@bhavin192 Addressed your comments.

bhavin192 (approved):

bhavin192:
@shinji62 have you tested these changes on a cluster? Just tried doing `helm template --set multicluster.createServicePerPod=true .` and it seems to be setting the rpc_bind_address to the service FQDN which won't work. 
Basically, just setting createServicePerPod with this change might not work. We might need to modify the rpc_bind_address to be $(POD_IP) I think? Need to test that. 
This template will need a change
```
{{- define "yugabyte.rpc_bind_address" -}}
```

shinji62:
@bhavin192 I did test before the first time I submit the PR, not every change, honestly I don't know much about what is expected for the rpc bind .. but I did the same as itio, pod ip or `127.0.0.1` etc ..

stable/yugabyte/templates/_helpers.tpl
```diff
@@ -243,7 +243,7 @@ we stick to 0.0.0.0, which works for master.
 */}}
 {{- define "yugabyte.rpc_bind_address" -}}
   {{- $port := index .Service.ports "tcp-rpc-port" -}}
-  {{- if .Values.istioCompatibility.enabled -}}
+  {{- if (or .Values.istioCompatibility.enabled .Values.multicluster.createServicePerPod) -}}
```
bhavin192: It would be better to set it to `$(POD_IP):{{ $port }}` i.e. same as createServiceExports. 
That seems to be a overall better option than 0.0.0.0.

bhavin192 (dismissed):

stable/yugabyte/templates/_helpers.tpl
shinji62: done @bhavin192

bhavin192 (approved):
Yatish has tested these changes. Only one thing to note here is when using the YB chart without YBA one has to set isMultiAZ=true, and set the masterAddresses to correct value to get this working. In case of YBA these are taken care of.

PR #166, opened 2023-05-08
[PLAT-8803] [YBA] Add extra volume support for postgres container

shinji62:
On a system with huge_page enabled it seems more reasonable to just mount the huge page config into the container, that will avoid any issues that we can faced when trying to deactivate huge_page which doesn't seems to fully work anyway.

For example for the Postgres container with we disabled huge_page on the Postgres conf the container start but resource shared_buffer  400Kb and max_connections 20 seems limited, 

```
fixing permissions on existing directory /var/lib/postgresql/data/pgdata ... ok  
creating subdirectories ... ok                          
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 20                
selecting default shared_buffers ... 400kB              
selecting default time zone ... Etc/UTC                 
creating configuration files ... ok                     
running bootstrap script ... ok                         
performing post-bootstrap initialization ... ok     
```

When mounting the hugepage into the container we can see the shared_buffer and the max connections are ok
```
fixing permissions on existing directory /var/lib/postgresql/data/pgdata ..ok            
creating subdirectories ... ok                           
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 100               
selecting default shared_buffers ... 128MB              
selecting default time zone ... Etc/UTC                 
creating configuration files ... ok                     
running bootstrap script ... ok                         
performing post-bootstrap initialization ... ok         
syncing data to disk ... ok
``` 

To reproduce 

1- Create a EKS cluster
2- SSH to one of the node and enable hugepagee
```
echo "vm.nr_hugepages=256" >> /etc/sysctl.conf
sysctl -p
sudo systemctl restart kubelet
```
3- Deploy yba to the specific host with node selector

```
nodeSelector:
  kubernetes.io/hostname: hostname # you can check with kubectl get nodes --show-labels
```

`helm upgrade --install  yw-test yugabytedb/yugaware  --version 2.17.2 -n yb-platform --values your_yba-value.yml`

4- if you don't mount the hugepage you will see that YBA can be deployed because of max connections reached.

### DIFF
--- a/stable/yugaware/templates/_helpers.tpl
+++ b/stable/yugaware/templates/_helpers.tpl
@@ -193,3 +193,41 @@ Check export of nss_wrapper environment variables required
       {{- printf "false" -}}
   {{- end -}}
 {{- end -}}
+
+
+{{/*
+  Verify the extraVolumes and extraVolumeMounts mappings.
+  Every extraVolumes should have extraVolumeMounts
+*/}}
+{{- define "yugaware.isExtraVolumesMappingExists" -}}
+  {{- $lenExtraVolumes := len .extraVolumes -}}
+  {{- $lenExtraVolumeMounts := len .extraVolumeMounts -}}
+
+  {{- if and (eq $lenExtraVolumeMounts 0) (gt $lenExtraVolumes 0) -}}
+    {{- fail "You have not provided the extraVolumeMounts for extraVolumes." -}}
+  {{- else if and (eq $lenExtraVolumes 0) (gt $lenExtraVolumeMounts 0) -}}
+    {{- fail "You have not provided the extraVolumes for extraVolumeMounts." -}}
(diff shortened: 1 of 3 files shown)


shinji62:
cc @iSignal

shinji62:
@bhavin192 Let me know if you need any changes

bhavin192 (approved):
The change looks good. I was able to test it with following YAML:

```yaml
postgres:
  extraVolumeMounts:
  - name: two
    mountPath: /some/tmp/
  extraVolumes:
  - name: two
    hostPath:
      path: /tmp
```

But seems like this needs a rebase, after that we can squash and merge this PR.

shinji62:
@bhavin192 I have updated the PR please merge it when you can, you will be nice to do the fix for 2.18 as well

PR #167, opened 2023-06-02
[Backport 2.18][PLAT-9120] [yugabyte] Decouple istio and servicePerPod

shinji62:
Backporting this

Decoupling service per pod and Istio
This allow to create service per pod without adding the side car injection, for example if I don't want my DB to be on my service mesh but I need to the service per pod because I would like to use XCluster without the need of pod-to-pod communication and without MCS.

Or in case of YBA been in a different namespace than YB and cross-namespace communication is not allow.

Test:
```
multicluster:
  createServicePerPod: true

tls:
  enabled: true

isMultiAz: true
masterAddresses: "yb-master-0.yb-platform.svc.cluster.local,yb-master-1.yb-platform.svc.cluster.local,yb-master-2.yb-platform.svc.cluster.local"
```
Both YSQL and YCQL work.
---------

### DIFF
--- a/stable/yugabyte/templates/_helpers.tpl
+++ b/stable/yugabyte/templates/_helpers.tpl
@@ -206,13 +206,13 @@ Get files from fs data directories for readiness / liveness probes.
 Generate server FQDN.
 */}}
 {{- define "yugabyte.server_fqdn" -}}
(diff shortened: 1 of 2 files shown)


bhavin192 (approved):
This has been tested with isMultiAZ and masterAddresses set.

PR #170, opened 2023-06-26
[yugabyte] Added YB controller port

shinji62:
Added YBC Port.
Port will not be added if YBC enabled is false.

### DIFF
--- a/stable/yugabyte/templates/multicluster/service-per-pod.yaml
+++ b/stable/yugabyte/templates/multicluster/service-per-pod.yaml
@@ -1,5 +1,4 @@
 {{- if .Values.multicluster.createServicePerPod }}
-{{- $root := . -}}
 {{- range $server := .Values.Services }}
 {{- range $replicaNum := until (int (ternary $.Values.replicas.master $.Values.replicas.tserver (eq $server.name "yb-masters"))) }}
 {{- $appLabelArgs := dict "label" $server.label "root" $ -}}
@@ -16,15 +15,15 @@ metadata:
 spec:
   ports:
     {{- range $label, $port := $server.ports }}
-      {{- if (eq $label "grpc-ybc-port")}}
-        {{- if (($root.Values.ybc.enabled))}}
+    {{- if (eq $label "grpc-ybc-port") }}
+    {{- if $.Values.ybc.enabled }}
     - name: {{ $label | quote }}
       port: {{ $port }}
-        {{- end}}
-      {{- else}}
+    {{- end }}
+    {{- else }}
     - name: {{ $label | quote }}
       port: {{ $port }}
-      {{- end}}
+    {{- end }}
     {{- end}}
   selector:
     statefulset.kubernetes.io/pod-name: {{ $podName | quote }}
--- a/stable/yugabyte/templates/multicluster/service-per-pod.yaml
+++ b/stable/yugabyte/templates/multicluster/service-per-pod.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.multicluster.createServicePerPod }}
+{{- $root := . -}}
 {{- range $server := .Values.Services }}
 {{- range $replicaNum := until (int (ternary $.Values.replicas.master $.Values.replicas.tserver (eq $server.name "yb-masters"))) }}
 {{- $appLabelArgs := dict "label" $server.label "root" $ -}}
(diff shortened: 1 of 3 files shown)


shinji62:
@bhavin192 Hope you can check that as well, my prospect may need the YBC to be exposed as well.

stable/yugabyte/templates/multicluster/service-per-pod.yaml
```diff
@@ -15,8 +16,15 @@ metadata:
 spec:
   ports:
     {{- range $label, $port := $server.ports }}
+      {{- if (eq $label "grpc-ybc-port")}}
+        {{- if (($root.Values.ybc.enabled))}}
```
bhavin192: You can just use `$` character instead of `$root` like we have done it on the line number 4. So, it will be 
```
$.Values.ybc.enabled
```

stable/yugabyte/templates/multicluster/service-per-pod.yaml
```diff
@@ -15,8 +16,15 @@ metadata:
...
     - name: {{ $label | quote }}
       port: {{ $port }}
+        {{- end}}
+      {{- else}}
+    - name: {{ $label | quote }}
+      port: {{ $port }}
+      {{- end}}
```
bhavin192:
```suggestion
    {{- if (eq $label "grpc-ybc-port") }}
    {{- if $.Values.ybc.enabled }}
    - name: {{ $label | quote }}
      port: {{ $port }}
    {{- end }}
    {{- else }}
    - name: {{ $label | quote }}
      port: {{ $port }}
    {{- end }}
```

How about keeping the indentation of template conditions same as YAML block, it is easy to add new YAML blocks that way.

stable/yugabyte/templates/service.yaml
```diff
@@ -92,8 +92,15 @@ spec:
...
     {{- range $label, $port := .ports }}
+      {{- if (eq $label "grpc-ybc-port")}}
+        {{- if (($root.Values.ybc.enabled))}}
     - name: {{ $label | quote }}
       port: {{ $port }}
+        {{- end}}
+      {{- else}}
+    - name: {{ $label | quote }}
+      port: {{ $port }}
+      {{- end}}
```
bhavin192:
```suggestion
    {{- if (eq $label "grpc-ybc-port") }}
    {{- if $root.Values.ybc.enabled }}
    - name: {{ $label | quote }}
      port: {{ $port }}
    {{- end }}
    {{- else }}
    - name: {{ $label | quote }}
      port: {{ $port }}
    {{- end }}
```

stable/yugabyte/templates/multicluster/service-per-pod.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if .Values.multicluster.createServicePerPod }}
+{{- $root := . -}}
```
bhavin192:
```suggestion
```

stable/yugabyte/templates/service.yaml
```diff
@@ -637,6 +646,10 @@ spec:
               sleep 60;
               /home/yugabyte/tools/k8s_ybc_parent.py status || /home/yugabyte/tools/k8s_ybc_parent.py start;
             done
+        ports:
+        {{- $ybcPort := index $service.ports "grpc-ybc-port" }}
+          - containerPort: {{ $ybcPort }}
+            name: "grpc-ybc-port"
```
bhavin192:
```suggestion
        {{- with index $service.ports "grpc-ybc-port" }}
        ports:
          - containerPort: {{ . }}
            name: "grpc-ybc-port"
        {{- end }}
```

We should use `with` here, as some of the users might have modified the `Services:` section. This section is an array which means these users will have whole section copied and added to the provider overrides. If we don't use with, their deployments will break. They won't have this new grpc-ybc-port field in their overrides.

This port is already exposed in case of normal users who use headless services. So, let's try not to break the installation for them.

bhavin192 (changes requested):
Thank you for working on this. I have added few suggestions, once those are addressed, we can merge this.

stable/yugabyte/templates/multicluster/service-per-pod.yaml
shinji62: I think much difficult to read but fine.

shinji62:
@bhavin192 thanks for the review I think everything have been addressed.

bhavin192 (approved):
Looks good! I tested following scenarios and it is working fine.
- ServicePerPod enabled, YBC enabled: each service has the port mentioned.
- YBC enabled, but `grpc-ybc-port` not in the list: port is not added, and installation is successful. This can happen when someone has overridden the Service section in their YBA configuration.
- YBC disabled: port is not added.

PR #174, opened 2023-09-26
Feature/add dnsconfig dnspolicy

shinji62:
Please check [IDEA-106](https://yugabyte.atlassian.net/jira/polaris/projects/IDEA/ideas/view/2813870?selectedIssue=IDEA-106) for details.

But in short, adding the feature to be able to specify dnsConfig / dnsPolicy

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -323,11 +323,11 @@ spec:
         {{- $_ := set $root.Values.tserver.affinity "podAntiAffinity" $podAntiAffinityData -}}
         {{ toYaml $root.Values.tserver.affinity | nindent 8 }}
         {{ end }}
-      {{- if $root.Values.dnsConfig }}
-      dnsConfig: {{- toYaml $root.Values.dnsConfig | nindent 8 }}
+      {{- with $root.Values.dnsConfig }}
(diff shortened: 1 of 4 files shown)


stable/yugabyte/templates/service.yaml
```diff
@@ -323,6 +323,12 @@ spec:
         {{- $_ := set $root.Values.tserver.affinity "podAntiAffinity" $podAntiAffinityData -}}
         {{ toYaml $root.Values.tserver.affinity | nindent 8 }}
         {{ end }}
+      {{- if $root.Values.dnsConfig }}
+      dnsConfig: {{- toYaml $root.Values.dnsConfig | nindent 8 }}
+      {{- end }}
```
bhavin192: We can use with here as well. Same for other occurrences. 
```suggestion
      {{- with $root.Values.dnsConfig }}
      dnsConfig: {{- toYaml . | nindent 8 }}
      {{- end }}
```

shinji62: Updated with changes @bhavin192

amannijhawan (approved):
Change looks good to me.
Not sure if we want to separate out dnsConfig of master and tserver but other than that change looks good.

shinji62:
@amannijhawan Same settings are ok for both master and tserver

PR #180, opened 2023-12-18
[#20344] yugabyted: yugabyted UI support for K8s helm chart

djiang9001:
Changes to the yugabyte helm chart to support yugabyted-ui:
- Enables yugabyted-ui by default on port 15433 for each tserver and master.
- Metrics snapshotter enabled by default since yugabyted-ui uses it to display metrics.
- Works with and without tls enabled (`tls.enabled=true` flag)
- Can be disabled with `yugabytedUi.enabled=false` flag)

Sample command:
```
helm install yb-demo /home/centos/code/charts-fork/stable/yugabyte \
--set resource.master.requests.cpu=0.5,resource.master.requests.memory=0.5Gi,\
resource.tserver.requests.cpu=0.5,resource.tserver.requests.memory=0.5Gi,\
replicas.master=3,replicas.tserver=3,tls.enabled=true --namespace yb-demo
```

**NOTE**: Do not merge until `tag` in `stable/yugabyte/values.yaml` has been updated to the correct value, as well as `version` and `appVersion` in `stable/yugabyte/Chart.yaml`.

### DIFF
--- a/stable/yugabyte/values.yaml
+++ b/stable/yugabyte/values.yaml
@@ -8,7 +8,7 @@ nameOverride: ""
 
 Image:
   repository: "yugabytedb/yugabyte"
-  tag: 2.19.3.0-b140
+  tag: latest
   pullPolicy: IfNotPresent
   pullSecretName: ""
 
@@ -115,6 +115,22 @@ gflags:
   tserver: {}
 #   use_cassandra_authentication: false
 
+yugabytedUi:
+  enabled: true
+
+metricsSnapshotter:
+  enabled: true
+  # time between each metric snapshot in ms
+  interval: 11000
+  whitelist:
+    - handler_latency_yb_tserver_TabletServerService_Read_count
+    - handler_latency_yb_tserver_TabletServerService_Write_count
+    - handler_latency_yb_tserver_TabletServerService_Read_sum
+    - handler_latency_yb_tserver_TabletServerService_Write_sum
+    - disk_usage
+    - cpu_usage
+    - node_up
+
 PodManagementPolicy: Parallel
 
 enableLoadBalancer: true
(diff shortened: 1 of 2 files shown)


stable/yugabyte/values.yaml
```diff
@@ -8,7 +8,7 @@ nameOverride: ""
 
 Image:
   repository: "yugabytedb/yugabyte"
-  tag: 2.19.3.0-b140
```
amannijhawan, line 11: please don't modify the values.yaml for this, this is modified by generated code.

stable/yugabyte/values.yaml
```diff
@@ -115,6 +115,22 @@ gflags:
   tserver: {}
 #   use_cassandra_authentication: false
 
+yugabytedUi:
+  enabled: true
```
amannijhawan, line 119: default should be false for yugabytedUI

stable/yugabyte/values.yaml
```diff
@@ -115,6 +115,22 @@ gflags:
...
+
+metricsSnapshotter:
```
amannijhawan, line 121: same here default should be false.

stable/yugabyte/values.yaml
```diff
@@ -504,6 +522,7 @@ helm2Legacy: false
 ip_version_support: "v4_only" # v4_only, v6_only are the only supported values at the moment
 
 # For more https://docs.yugabyte.com/latest/reference/configuration/yugabyted/#environment-variables
+# Note: for yugabyted-ui to work properly with credentials, ysql/ycql passwords must be the same
```
amannijhawan, line 525: Why is this the case? 
This can be a blocker for the operator integration?

amannijhawan (changes requested):

amannijhawan:
Hi @djiang9001  
I did a first round review but can you please publish this via phorge/phabricator pipeline. 
Also can you please add rendered helm charts as well as a link to where this is working in the Testing Done section ?
Thanks
Aman

PR #181, opened 2024-01-10
Use HTTPS to scrape platform metrics when tls is enabled.

shinji62:
PLAT-12210 When TLS is enabled for YBA Yugaware we should use https to scrape the platform metrics endpoint.

Added some empty line just for rendering.

Tested on EKS

Only unknow is Nginx, not sure we should still support it or not and what need to be done to support it.

### DIFF
--- a/stable/yugaware/templates/configs.yaml
+++ b/stable/yugaware/templates/configs.yaml
@@ -583,17 +583,8 @@ data:
       {{- end }}
       {{- end }}
  
-      {{- if not .Values.tls.enabled }}
-
-      - job_name: 'platform'
-        metrics_path: "/api/v1/prometheus_metrics"
-        static_configs:
-          - targets: [
-            '{{ eq .Values.ip_version_support "v6_only" | ternary "[::1]" "127.0.0.1" }}:9000'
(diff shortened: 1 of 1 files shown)


amannijhawan (approved):

stable/yugaware/templates/configs.yaml
```diff
@@ -582,6 +582,8 @@ data:
 
       {{- end }}
       {{- end }}
+ 
+      {{- if not .Values.tls.enabled }}
```
bhavin192: I think we should also use this block if `useNginxProxy` is enabled. 

Basically, if someone is using nginx proxy, then the YBA will be listening on 9000 (http).

bhavin192 (commented):
Looks good overall, only one comment related to Nginx, as it is still there as an option and not completely removed yet.

stable/yugaware/templates/configs.yaml
shinji62: Please check @bhavin192 , I have addressed your comment.

bhavin192 (approved):

PR #185, opened 2024-02-19
Add startup and liveness probes to yugaware container

skorobogatydmitry:
PR adds 2 health probes for yugaware container within a pod.
Startup probe has low `initialDelaySeconds`, what allows the pod to be quicky available for small deployments. High `failureThreshold` allows to tolerate long-running DB migrations and other startup operations. 10s `periodSeconds` ensures a quick reaction on container state change.
Liveness probe checks container less frequently, but tolerates only 3 failed probes.

### DIFF
--- a/stable/yugaware/templates/statefulset.yaml
+++ b/stable/yugaware/templates/statefulset.yaml
@@ -538,6 +538,21 @@ spec:
           ports:
             - containerPort: 9000
               name: yugaware
+          startupProbe: # startup could take from ~30 seconds to ~5 minutes
+            failureThreshold: 60
+            initialDelaySeconds: 20
+            periodSeconds: 10
+            timeoutSeconds: 5
+            httpGet:
+                path: /api/app_version
+                port: yugaware
+          livenessProbe: # less frequent but more sensitive than startupProbe
+            failureThreshold: 3
+            periodSeconds: 30
+            timeoutSeconds: 10
+            httpGet:
+                path: /api/app_version
+                port: yugaware
           volumeMounts:
           - name: yugaware-config
             mountPath: /data
--- a/stable/yugaware/templates/statefulset.yaml
+++ b/stable/yugaware/templates/statefulset.yaml
@@ -538,21 +538,23 @@ spec:
           ports:
             - containerPort: 9000
               name: yugaware
-          startupProbe: # startup could take from ~30 seconds to ~5 minutes
-            failureThreshold: 60
-            initialDelaySeconds: 20
-            periodSeconds: 10
-            timeoutSeconds: 5
+          {{- if .Values.yugaware.pod.probes.enabled }}
+          startupProbe:
+            failureThreshold: {{ div .Values.yugaware.pod.probes.startupTimeSec.max (div .Values.yugaware.pod.probes.startupTimeSec.min 2) }}
+            initialDelaySeconds: {{ .Values.yugaware.pod.probes.startupTimeSec.min }}
+            periodSeconds: {{ div .Values.yugaware.pod.probes.startupTimeSec.min 2 }}
+            timeoutSeconds: {{ div .Values.yugaware.pod.probes.startupTimeSec.min 4 }}
             httpGet:
                 path: /api/app_version
                 port: yugaware
-          livenessProbe: # less frequent but more sensitive than startupProbe
-            failureThreshold: 3
-            periodSeconds: 30
-            timeoutSeconds: 10
+          livenessProbe:
+            failureThreshold: {{ .Values.yugaware.pod.probes.livenessProbe.failureThreshold }}
+            periodSeconds: {{ .Values.yugaware.pod.probes.livenessProbe.periodSeconds }}
+            timeoutSeconds: {{ div .Values.yugaware.pod.probes.livenessProbe.periodSeconds 2 }}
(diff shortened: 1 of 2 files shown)


stable/yugaware/values.yaml
```diff
@@ -84,6 +84,14 @@ yugaware:
...
     labels: {}
     terminationGracePeriodSeconds: 30
+    probes: # HTTP healthcheck probes configuration
+      enabled: false # add or not livenessProbe and startupProbe to the container
+      startupTimeSec:
+        min: 20 # minimum expected time for yugaware to start
+        max: 600 # maximum expected time for yugaware to start, k8s will restart pod on timeout
+      livenessProbe:
+        periodSeconds: 30 # query period
+        failureThreshold: 3 # number of failed checks to restart pod
```
amannijhawan: can we bump this up in default to something like 10 failed probes? 
YBM can set a more agressive threshold if needed?

stable/yugaware/templates/statefulset.yaml
```diff
@@ -538,6 +538,23 @@ spec:
...
+            initialDelaySeconds: {{ .Values.yugaware.pod.probes.startupTimeSec.min }}
+            periodSeconds: {{ div .Values.yugaware.pod.probes.startupTimeSec.min 2 }}
+            timeoutSeconds: {{ div .Values.yugaware.pod.probes.startupTimeSec.min 4 }}
+            httpGet:
+                path: /api/app_version
+                port: yugaware
+          livenessProbe:
+            failureThreshold: {{ .Values.yugaware.pod.probes.livenessProbe.failureThreshold }}
+            periodSeconds: {{ .Values.yugaware.pod.probes.livenessProbe.periodSeconds }}
+            timeoutSeconds: {{ div .Values.yugaware.pod.probes.livenessProbe.periodSeconds 2 }}
```
amannijhawan, line 553: is this a connection timeout for the probe ? 
Can you describe why it should be 0.5*periodSeconds?

stable/yugaware/templates/statefulset.yaml
```diff
@@ -538,6 +538,23 @@ spec:
           ports:
             - containerPort: 9000
               name: yugaware
+          {{- if .Values.yugaware.pod.probes.enabled }}
+          startupProbe:
+            failureThreshold: {{ div .Values.yugaware.pod.probes.startupTimeSec.max (div .Values.yugaware.pod.probes.startupTimeSec.min 2) }}
+            initialDelaySeconds: {{ .Values.yugaware.pod.probes.startupTimeSec.min }}
+            periodSeconds: {{ div .Values.yugaware.pod.probes.startupTimeSec.min 2 }}
+            timeoutSeconds: {{ div .Values.yugaware.pod.probes.startupTimeSec.min 4 }}
+            httpGet:
```
amannijhawan, line 547: Can you describe the logic behind the derived values? 
Do we need to derive these values? Can we not have them defined separately in values.yaml?

stable/yugaware/values.yaml
skorobogatydmitry: Yes, will do.
Although HTTP being unavailable for 5 minutes seems to relaxed to me. It usually means there are perfomance issues.

stable/yugaware/templates/statefulset.yaml
skorobogatydmitry: Disclaimer: I didn't check how k8s probes works exactly.
Timeout affects probes period, as it reflectx maximum time a probe could take.
If a probe time is taken into account to ensure period is static, it should be less than period to avoid accumulating probes. Divisinon by 2 looks fine for this case and gives some space for other hiccups.
In case a probe time isn't counted as a part of period, it effectively extends period. Extending by 50% is something this workload could live with.

stable/yugaware/templates/statefulset.yaml
skorobogatydmitry: I converted an awkward k8s setting to something meaningful for startup probe parameters.
Startup probe suites to tolerate long application startups before enabling liveness and readiness probes.
From that perspective, 2 numbers are interesting: minimum time app startup typically takes and maximum time, as we want an app to be available asap, but wait long enough to swallow long startups. _k8s doesn't add pod to service up until it passes startup probe in absense of liveness_
Minimum time is put to `initialDelaySeconds`. First probe happens after this delay.
In case of YBA, minimum startup time divided by two makes a good fit for `periodSeconds`. E.g. 30 sec initial delay ~ 15 sec period, 20 sec delay ~ 10 sec period, etc
In order to wait as long as is in `max`, we need to could # of periods in `max`. That's what `{{ div .Values.yugaware.pod.probes.startupTimeSec.max (div .Values.yugaware.pod.probes.startupTimeSec.min 2) }}` does.

Do you want me to dump this to the chart itself ?

amannijhawan: Thanks for the explaination, i think adding this to commit message would be good enough.

amannijhawan (approved):
Change looks fine, especially with enabled: false by default. 
Is there any way to catch the failure of readiness probes in logs  or events that YBM monitors.
It would be good to catch these issues if they happen in YBM production or enhance the context of bug reports in which we see them so we know why YBA restarted or why an ongoing task was interrupted.

PR #191, opened 2024-10-10
[CLOUDGA-24038] Make persistent volume optional

nadarashwin:
Make persistent volume optional to make the underlying platform volume ephemeral.

### DIFF
--- a/stable/yugaware/values.yaml
+++ b/stable/yugaware/values.yaml
@@ -66,6 +66,9 @@ yugaware:
   storageClass: ""
   storageAnnotations: {}
   multiTenant: false
+  ## This is for testing purposes only.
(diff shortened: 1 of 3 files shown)


rtsisyk (approved):

stable/yugaware/values.yaml
```diff
@@ -66,6 +66,8 @@ yugaware:
   storageClass: ""
   storageAnnotations: {}
   multiTenant: false
+  persistentVolume:
+    enabled: true
```
iSignal, line 71: Can you comment that this is for testing purposes only

iSignal (approved):

amannijhawan (approved):

Issue #194, opened 2024-11-08
Optionally use secret to set env variables in setup credentials job


LilMonk:
Give an option to use a kubernetes secret for setting these env variables.
https://github.com/yugabyte/charts/blob/8c8acc0c66900cb47d90ee8aa300d07776956ea4/stable/yugabyte/templates/hooks/setup-credentials-job.yaml#L32-L51

amannijhawan:
Thanks @LilMonk, we will look at fixing this with the next release.

PR #197, opened 2025-03-10
Optionally use secret to set env variables in setup credentials job #194

shubin-yb:
Allow using secrets to setup ysql/ycql passwords. In addition, created some unit tests to validate the output, and unit tests will now run as part of pull requests


shubin-yb:
merged changes using internal Yugabyte tools

PR #203, opened 2025-04-30
Add 'provided' flag for tls certificates

the-glu:
This PR add a 'provided' flag in TLS settings.

The goal of this flag is to disable secrets handling from parameters, but let users of the chart manage those secrets as needed/wanted.

It's doesn't change anything by default and it backward compatible :)

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -1,6 +1,6 @@
 {{- $root := . -}}
 ---
-{{- if and (eq $root.Values.tls.enabled true) (eq $root.Values.tls.certManager.enabled false) }}
+{{- if and (eq $root.Values.tls.enabled true) (eq $root.Values.tls.certManager.enabled false) (eq $root.Values.tls.provided false) }}
 ---
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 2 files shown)


barroco:
@amannijhawan @shubin-yb @daniel-yb  thanks for the great work on this repository. We are waiting for this change in the [Linux Foundation InterUSS project](https://github.com/interuss/dss) to progress on the migration to Yugabyte. May I kindly ask you for a ETA to review this change ? Please, don't hesitate to let us know if there is anything we could do to simplify your work.

iSignal:
@barroco : @the-glu : Thanks for the context.

Is the goal to generate per-pod certs for master/tserver outside the helm charts and pre-create them with the names expected in the helm chart?

Is cert-manager an option here? That would be the preferred workflow for such use cases.

Without cert-manager, the current option is a bit clunky as it would require a user to first guess the names of the pods being created and the appropriate secret files required before setting up TLS. At the very least, we should add some context to this PR describing all the secrets that need to be pre-created for this option to work.

the-glu:
> Is the goal to generate per-pod certs for master/tserver outside the helm charts and pre-create them with the names expected in the helm chart?

The goal is to have per-pod certs for master/tserver outside with the known in advances, public dns name that we would like to use to form a cluster. We also include the names expected in the helm chart since they are known in advance. We also would like to have a set of known CA, not just one (that we do share via other means).

> Is cert-manager an option here? That would be the preferred workflow for such use cases.

Not sure it would help in that case, especially with the multiples CA ^^'

> Without cert-manager, the current option is a bit clunky as it would require a user to first guess the names of the pods being created and the appropriate secret files required before setting up TLS. At the very least, we should add some context to this PR describing all the secrets that need to be pre-created for this option to work.

Yes, but those names are known in advance (public ones and internal ones) since they are in StatefulSet :)

I will add doc about the needed secret

amannijhawan:
Will we be directly modifying the gflags in this case to send the secrets in ? 
Just want to understand how this is going to be used to be able to test this.

the-glu:
> Will we be directly modifying the gflags in this case to send the secrets in ? Just want to understand how this is going to be used to be able to test this.

Hum no: it's just disabling certificate management in the chart, but still get them as before from the secrets. Someone using this should manage the secrets as expect by the chart :)

PR #207, opened 2025-05-27
Add more overrides for rpc bind address and server broadcast address

the-glu:
This PR adds optional options to overrides the rpc bind address on master and tserver, and server_broadcast_address on the master (already present on the tserver).

### DIFF
--- a/stable/yugabyte/values.yaml
+++ b/stable/yugabyte/values.yaml
@@ -457,6 +457,15 @@ master:
   #       fieldPath: status.hostIP
   extraEnv: []
 
+  ## Sets the --server_broadcast_addresses flag on the Master pods, no
+  ## preflight checks are done for this address. You might need to add
+  ## `use_private_ip: cloud` to the gflags.master and gflags.tserver.
+  serverBroadcastAddress: ""
+
+  ## Sets the --rpc_bind_addresses flag on the Master pods, no
+  ## preflight checks are done for this address.
+  rpcBindAddress: ""
+
   # secretEnv variables are used to expose secrets data as env variables in the master pod.
   # TODO Add namespace also to support copying secrets from other namespace.
   # secretEnv:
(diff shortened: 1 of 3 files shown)


kv83821-yb:
Hi @the-glu 
Can you please add some more context on this ? 
Right now, we set the rpc_bind_address as the POD_IP except for Multi-Cluster cases.
The broadcast address is set to pod fqdn. Are you trying to work with any specific MCS setup ?

the-glu:
Hello,

> Right now, we set the rpc_bind_address as the POD_IP except for Multi-Cluster cases. The broadcast address is set to pod fqdn. Are you trying to work with any specific MCS setup ?

Yes, I'm trying to do a multi cluster setup and yugaybe is using the private_rpc_address even with use_public_ip=never: https://github.com/yugabyte/yugabyte-db/issues/27367

With thoses I'm able to work around the issue by forcing the private_rpc_address to a public one :)

kv83821-yb:
Which mcs flavor is this ? we can perhaps try to add support for that itself instead of providing direct overrides ? cc: @amannijhawan 

Right now we support gke, istio, ocp mcs

the-glu:
I did try with gke and eks, but I'm not sure it's directly related to the provider

stable/yugabyte/values.yaml
```diff
@@ -633,6 +642,10 @@ tserver:
   ## `use_private_ip: cloud` to the gflags.master and gflags.tserver.
   serverBroadcastAddress: ""
```
amannijhawan, line 679: How will this work in case of multiple masters? Are we mostly setting it to 0.0.0.0?

amannijhawan:
Added a comment, How will this work if we have more than 1 master deployed with the chart  ?

stable/yugabyte/values.yaml
the-glu: It's a override in case of someone would like to update it. Without anything set, it will use the `server_broadcast_address` template, who include [the server fqdn](https://github.com/yugabyte/charts/blob/04a27b8f42f816951d8e71adda5c00977b2903ef/stable/yugabyte/templates/_helpers.tpl#L288)

PR #208, opened 2025-05-27
Add pre commands options

the-glu:
This PR add the option to run commands before the initialization of the pod.

### DIFF
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -405,7 +405,7 @@ spec:
           - "/bin/bash"
           - "-c"
           - |
-          {{- $preCommands := (ternary $root.Values.master.advanced.preCommands $root.Values.tserver.advanced.preCommands (eq .name "yb-masters")) }} {{- with $preCommands }} {{ . | nindent 12 }} {{- end }}
+          {{- $preCommands := ( (eq .name "yb-masters") | ternary $root.Values.master.advanced.preCommands $root.Values.tserver.advanced.preCommands ) }} {{- with $preCommands }} {{ . | nindent 12 }} {{- end }}
           {{- if and (not $root.Values.preflight.skipUlimit) (not $root.Values.preflight.skipAll) }}
             if [ -f /home/yugabyte/tools/k8s_preflight.py ]; then
               /home/yugabyte/tools/k8s_preflight.py all
--- a/stable/yugabyte/templates/service.yaml
+++ b/stable/yugabyte/templates/service.yaml
@@ -405,6 +405,7 @@ spec:
           - "/bin/bash"
           - "-c"
           - |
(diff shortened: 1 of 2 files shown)


Arpit-yb:
My primary concern here revolves around the potential security/supportability implications of allowing an unrestricted preCommands field for yb-masters and tservers. While the flexibility is understood, this approach currently grants significant power within the pod. Will let @amannijhawan or @subramanian-neelakantan  comment on that.

NIT: It would be better to use a ternary operator instead of if statements for conciseness. Something like this should work well:
`{{- $preCommands := (ternary $root.Values.master.preCommands $root.Values.tserver.preCommands (eq .name "yb-masters")) }}
{{- with $preCommands }}
  {{ . | nindent 12 }}
{{- end }}`

Arpit-yb:
We discussed this internally and are good to proceed with it. Could you move the preCommand option from master and tserver into a new advanced section?

The structure should look like this:

```
master:
    advanced:
        preCommand: ""
```
Same for tserver as well

the-glu:
Notice I need the flexibility as a workaround for https://github.com/yugabyte/yugabyte-db/issues/27367, so I would understand that security is more important that allowing random commands ;)

I fixed PR with the ternary and advanced section :)

Arpit-yb (approved):
LGTM Thanks. 
NIT - Please update the comment with a warning. Maybe something like - 

```
#   WARNING: This is an advanced feature. Incorrect use may cause startup failures,
#   security issues, or unexpected behavior. Use with caution and at your own risk.
```

the-glu:
> NIT - Please update the comment with a warning. Maybe something like -

Done in https://github.com/yugabyte/charts/pull/208/commits/cfcaa4da249fff1ea2e17ea99ee004b1210eab86 :)

stable/yugabyte/templates/service.yaml
```diff
@@ -405,6 +405,7 @@ spec:
           - "/bin/bash"
           - "-c"
           - |
+          {{- $preCommands := (ternary $root.Values.master.advanced.preCommands $root.Values.tserver.advanced.preCommands (eq .name "yb-masters")) }} {{- with $preCommands }} {{ . | nindent 12 }} {{- end }}
```
kv83821-yb: Nit: most of our charts use
`true | <branch true> <branch false>` syntax for ternary.

kv83821-yb (approved):
Lgtm! Please see if you can resolve nit

stable/yugabyte/templates/service.yaml
the-glu: Fixed :)

the-glu:
> Lgtm! Please see if you can resolve nit

Yes, done :)

the-glu:
@kv83821-yb @Arpit-yb could this one be merged?

Arpit-yb:
Yes, we can merge it. Thanks!

PR #212, opened 2025-10-02
[CLOUDGA-29511] Make promethues container optional

nadarashwin:
Different deployments employ distinct metrics collection approaches.
Some utilize Prometheus sidecar containers, while others leverage PodMonitor custom resources or Kubernetes service discovery mechanisms.
So we make the Prometheus container optional.


anmalysh-yb (approved):
LGTM.

amannijhawan:
Can you get the logs of the deployed YBA, YBA will constantly try to connect to prometheus so there will be large number of stack traces here.

PR #228, opened 2026-09-09
Charts: Allow exporting serviceEndpoints services as ServiceExport for MCS

kv83821-yb:
## Why

This is needed for the **namespace sameness** case of MCS, where clusterset-based DNS has to be able to reach all the pods across the clusters.

Today `multicluster.createServiceExports` only exports the headless `yb-masters` / `yb-tservers` services. The services defined under `serviceEndpoints` (`yb-master-ui`, `yb-tserver-service`, `yugabyted-ui-service`, ...) stay cluster-local, so they are not resolvable over `*.svc.clusterset.local` from the peer clusters of the clusterset. This adds an opt-in way to export them too.

## What

A per-endpoint `exportService` flag on `serviceEndpoints`, **defaulting to `false`** so existing installs are unaffected:

```yaml
serviceEndpoints:
  - name: "yb-tserver-service"
    type: ClusterIP
    scope: "Namespaced"
    app: "yb-tserver"
    ## Set to true to create a ServiceExport (KEP-1645) for this
    ## service, making it reachable from the other clusters of the
    ## ClusterSet.
    exportService: true
    ports:
      tcp-ysql-port: "5433"
```

When the flag is set, a `ServiceExport` is emitted right after that endpoint's `Service` in `templates/service-endpoints.yaml`:

- It sits inside the same render guards, so it inherits the existing `enableLoadBalancer`, app-match, naming-style, scope and `yugabyted-ui` conditions — no export is ever created without its Service.
- It reuses the same `$serviceName`, so the export always matches the name of the Service it belongs to, under either naming style.
- It carries the same labels as the Service, and gets `helm.sh/resource-policy: keep` for `Namespaced` scope, matching what the Service already does.
- `apiVersion` comes from the existing `multicluster.mcsApiVersion`, so `net.gke.io/v1` works for GKE MCS.
- This is independent of `multicluster.createServiceExports`, which keeps exporting only the headless services.

Endpoint-level `annotations` are deliberately **not** copied onto the export — those are load-balancer annotations and carry no meaning on a `ServiceExport`.

## Testing

New suite `stable/yugabyte/tests/test_service_endpoint_export.yaml` covers: nothing exported by default, only opted-in endpoints exported, `mcsApiVersion` + new-naming-style naming, and the namespaced `keep` annotation.

```
$ cd stable/yugabyte && helm unittest -f "tests/test_*.yaml" .
Test Suites: 8 passed, 8 total
Tests:       57 passed, 57 total
```

`helm lint` is clean apart from the two pre-existing `policy/v1beta1` PodDisruptionBudget deprecation warnings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/stable/yugabyte/templates/service-endpoints.yaml
+++ b/stable/yugabyte/templates/service-endpoints.yaml
@@ -84,6 +84,34 @@ spec:
   {{ $key }}: {{ $value | quote }}
   {{- end }}
 ---
+{{- /* Export this endpoint service to the ClusterSet (KEP-1645) */}}
+{{- if $endpoint.exportService }}
+apiVersion: {{ $root.Values.multicluster.mcsApiVersion }}
+kind: ServiceExport
+metadata:
+  name: {{ $serviceName | quote }}
+  namespace: "{{ $root.Release.Namespace }}"
+  {{- if eq $scope "Namespaced" }}
+  annotations:
+    # Keep resource for namespaced services
+    "helm.sh/resource-policy": keep
+  {{- end }}
+  labels:
+    # scope is "Namespaced" or "AZ"
+    scope: {{ $scope }}
+    serviceName: {{ $endpoint.name }}
+    service-type: "endpoint"
+    {{- include "yugabyte.applabel" ($appLabelArgs) | indent 4 }}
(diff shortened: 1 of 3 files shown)


amannijhawan:
LGTM

amannijhawan (approved):

